PR #9, opened 2026-07-16
Fix launcher checks behind localhost proxies

### LINKED ISSUE #8: [Bug bounty] monk-agent launcher routes localhost health checks through configured proxy
### Stage

install / connect

### Coding agent

OpenAI Codex

### Coding agent version

codex-cli 0.144.5; Monk plugin 0.1.42

### Repro steps

1. On macOS, run a local HTTP proxy and export proxy variables without a localhost exception. For example:

```text
http_proxy=http://127.0.0.1:10808
https_proxy=http://127.0.0.1:10808
all_proxy=socks5://127.0.0.1:10808
```

2. Start `monk-agent` normally on `127.0.0.1:7419`.
3. Verify that the companion is healthy when contacted directly:

```text
curl --noproxy 127.0.0.1 http://127.0.0.1:7419/.well-known/oauth-protected-resource
HTTP/1.1 200 OK
{"resource":"http://127.0.0.1:7419/mcp", ...}
```

4. Run the plugin launcher, or request the same health endpoint without a localhost proxy bypass. The current launcher uses ordinary `curl`/`wget` calls for the health check, auth-status check, and sign-in nudge.
5. Observe that `curl` uses `http_proxy`, connects to the proxy at `127.0.0.1:10808`, and receives `HTTP/1.1 503 Service Unavailable` instead of contacting the healthy agent directly.

### Expected behavior

All launcher requests to the loopback-only Monk companion should bypass user-configured HTTP, HTTPS, and ALL_PROXY settings. A healthy local agent should be detected as running, and the auth-status and sign-in nudge requests should reach that agent.

### Actual behavior

The localhost health request is routed through the configured proxy:

```text
Uses proxy env variable http_proxy == 'http://127.0.0.1:10808'
Connected to 127.0.0.1 port 10808
GET http://127.0.0.1:7419/.well-known/oauth-protected-resource
HTTP/1.1 503 Service Unavailable
```

The same endpoint returns `200 OK` immediately with `curl --noproxy 127.0.0.1`. As a result, the launcher misdiagnoses a healthy `monk-agent` as unavailable and may attempt unnecessary restart/install recovery. The auth-status and sign-in nudge calls have the same proxy-routing problem.

### Severity (your guess)

blocker

### OS

macOS 26.5.1 (25F80), Apple Silicon

### monkd version

Monk CLI v3.21.0, build c93b59a9; local monk-agent on 127.0.0.1:7419

### Target cloud

Local Monk runtime (before target-cloud deployment)

### Integration (if relevant)

Monk plugin in OpenAI Codex

### Proposed fix

Explicitly bypass the configured proxy for local companion requests: use `curl --noproxy "$host"`, and set both `NO_PROXY` and `no_proxy` for the `wget` fallback. Apply it consistently to the health probe, auth-status probe, and sign-in nudge. A tested patch is ready.

gebibd00-jpg:
## Summary

- bypass user-configured proxies for loopback `monk-agent` health and auth checks
- apply the bypass to the sign-in nudge request as well
- cover both `curl` and the `wget` fallback in both distributed launcher script copies

## Root cause

The launcher inherited `http_proxy`, `https_proxy`, and `ALL_PROXY` settings for requests to the loopback-only companion. When localhost was not present in `NO_PROXY`, those requests were sent to the user's proxy instead of `monk-agent`. A proxy response such as HTTP 503 then made the launcher misdiagnose a healthy local agent as unavailable.

## User impact

Users running a local proxy can now start and reuse `monk-agent` normally. Health detection, auth-state detection, and the sign-in nudge reach the local companion directly without requiring users to change their global proxy configuration.

## Validation

- reproduced the issue against a live companion on `127.0.0.1:7419`: proxied request returned HTTP 503 while the direct request returned HTTP 200 and the expected resource document
- `sh -n scripts/start-monk-agent.sh`
- `sh -n plugins/monk/scripts/start-monk-agent.sh`
- `git diff --check 19b10e3^ 19b10e3`

Fixes #8

### DIFF
--- a/plugins/monk/scripts/start-monk-agent.sh
+++ b/plugins/monk/scripts/start-monk-agent.sh
@@ -45,6 +45,13 @@ mkdir -p "$log_dir" "$run_dir"
 
 health_url="http://$host:$port/.well-known/oauth-protected-resource"
 
+# Requests to the local companion must never be sent through a user-configured
+# HTTP(S) or ALL_PROXY. A proxy cannot normally reach the loopback-only agent
+# and may return its own response, which makes the launcher misdiagnose a
+# healthy agent as unavailable. curl gets an explicit bypass; wget uses the
+# standard no_proxy variables.
+agent_no_proxy="$host,localhost"
+
 # Register monk in Antigravity's global MCP config if ~/.gemini/config/ exists.
 # Idempotent — skips if already registered. Uses jq when available; prints
 # manual instructions otherwise.
@@ -117,9 +124,9 @@ emit_signin_nudge() {
   attempt=0
   while [ "$attempt" -lt 3 ]; do
     if command -v curl >/dev/null 2>&1; then
-      body="$(curl -fsS --max-time 5 "$status_url" 2>/dev/null || true)"
+      body="$(curl --noproxy "$host" -fsS --max-time 5 "$status_url" 2>/dev/null || true)"
     elif command -v wget >/dev/null 2>&1; then
-      body="$(wget -q -T 5 -O - "$status_url" 2>/dev/null || true)"
+      body="$(NO_PROXY="$agent_no_proxy,${NO_PROXY:-}" no_proxy="$agent_no_proxy,${no_proxy:-}" wget -q -T 5 -O - "$status_url" 2>/dev/null || true)"
     fi
     [ -n "$body" ] && break
     attempt=$((attempt + 1))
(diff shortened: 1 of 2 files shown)


nooga:
Thanks for the fix. On review this only updates 2 of the 3 rendered copies of the launcher script (misses one), so it's not a complete fix yet. Closing for now -- happy to look again if updated to cover all copies.

PR #11, opened 2026-07-16
Reuse healthy custom monk-agent paths

### LINKED ISSUE #10: [Bug bounty] custom MONK_AGENT_PATH restarts a healthy companion every session
### Stage

connect

### Coding agent

OpenAI Codex

### Coding agent version

codex-cli 0.144.5

### Repro steps

### Expected behavior

### Actual behavior

### Severity (your guess)

major

### OS

### monkd version

### Target cloud

Local companion only

### Integration (if relevant)

Monk plugin in OpenAI Codex

Dillxn:
## Summary

- calculate binary-update hashes only when the managed installer actually ran
- record the selected custom executable path and reuse a healthy companion only when that path still matches
- make the macOS launchd configuration check include the selected executable
- apply the same behavior to the POSIX and PowerShell launchers across all three distributed script copies
- add deterministic regression coverage for reuse when the custom path is unchanged and one restart when it changes

## Root cause

With `MONK_AGENT_PATH` set, the launchers calculated the “before” hash from the managed install but the “after” hash from the custom executable. Those two unrelated files normally differ, so every session set `agent_updated`/`AgentUpdated` and bypassed the healthy-agent fast path.

## User impact

A custom companion now restarts once when the configured executable changes, then remains running across later sessions while it is healthy. This avoids unnecessary process or launchd churn and prevents avoidable MCP/auth-session interruption during local plugin development.

## Validation

- reproduced the failure against the unmodified launcher with a healthy local endpoint fixture
- `sh tests/start-monk-agent-custom-path.sh` (proves unchanged-path reuse and changed-path restart)
- `sh -n scripts/start-monk-agent.sh`
- `sh -n plugins/monk/scripts/start-monk-agent.sh`
- confirmed root and packaged POSIX launchers are identical
- confirmed root and packaged PowerShell launchers are identical
- confirmed the hidden Antigravity launcher copies are identical too
- `git diff --check`

Fixes #10

### DIFF
--- a/.antigravity-plugin/scripts/start-monk-agent.ps1
+++ b/.antigravity-plugin/scripts/start-monk-agent.ps1
@@ -18,6 +18,7 @@ $RunDir = Join-Path $DataDir "run"
 $LogOut = Join-Path $LogDir "monk-agent.out.log"
 $LogErr = Join-Path $LogDir "monk-agent.err.log"
 $PidFile = Join-Path $RunDir "monk-agent.pid"
+$PathFile = Join-Path $RunDir "monk-agent.path"
 $HealthUrl = "http://${AgentHost}:$Port/.well-known/oauth-protected-resource"
 
 New-Item -ItemType Directory -Force -Path $LogDir, $RunDir | Out-Null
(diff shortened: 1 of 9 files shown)


Dillxn:
Maintainer note for the active plugin hunt: this PR remains intentionally scoped to #10 and includes a deterministic regression for unchanged-path reuse plus exactly one restart when the custom path changes. I have separate tested fixes for #19, #21, and #25, but I am keeping them out of this PR to preserve the one-bug-per-issue review boundary. Once this PR clears, I can submit those fixes sequentially without mixing causes. Happy to adjust this patch if you want a narrower or different integration.

Dillxn:
Synced this PR with current upstream `v0.1.45` and revalidated the merged result today.

Validation:
- `sh tests/start-monk-agent-custom-path.sh` — passes both unchanged-path reuse and one restart after a path switch
- all three rendered POSIX launcher copies are byte-identical
- all three rendered PowerShell launcher copies are byte-identical
- `sh -n` passes for every POSIX copy
- `git diff --check upstream/main...main` passes
- fork Install E2E passed on Ubuntu, macOS, and Windows (including native PowerShell install/status/start): https://github.com/Dillxn/monk-plugin/actions/runs/29800523406

The scope remains only #10; no unrelated fixes were added.

Lachann:
Thanks for this — your fix (managed-vs-custom hash comparison) was ported into the fix that shipped in plugin release **v0.1.50**. Closing without merging directly since `main` has since diverged (combined with related readiness/fastpath fixes in the same files), but full credit goes to this PR.

PR #13, opened 2026-07-16
fix(launcher): don't abort on malformed Antigravity MCP config

### LINKED ISSUE #12: [Bug bounty] malformed Antigravity config aborts a healthy launcher
### Stage

connect

### Coding agent

OpenAI Codex

### Coding agent version

codex-cli 0.144.5

### Repro steps

1. Have Antigravity's config directory present at `~/.gemini/config/`.
2. Put malformed, non-empty JSON in `~/.gemini/config/mcp_config.json`, such as `{invalid`.
3. Start or reuse `monk-agent` through `scripts/start-monk-agent.sh` on a machine with `jq` installed.
4. Let the normal launcher health check succeed.

I reproduced this deterministically with a healthy companion endpoint. The launcher exits with status 5 and prints:

```text
jq: parse error: Invalid numeric literal
```

### Expected behavior

Automatic Antigravity registration is optional and documented as best-effort. An unreadable or malformed unrelated config should produce a warning or manual registration instructions, then allow the healthy Monk companion and the current host session to continue.

### Actual behavior

`register_antigravity_mcp` runs `jq` without handling parse or write failure. Because the launcher uses `set -e`, `jq`'s non-zero status terminates the entire launcher after the Monk health check has already passed.

The registration runs whenever `~/.gemini/config/` exists, even when the current host is Codex or Claude Code. As a result, a damaged Antigravity config can prevent an otherwise healthy Monk plugin session from starting in another coding agent.

### Severity (your guess)

major

### OS

macOS 26.5.2 (25F84), Apple Silicon

### monkd version

Not applicable; the failure is in the plugin launcher after the companion health check

### Target cloud

Local companion only

### Integration (if relevant)

Monk plugin in OpenAI Codex; optional Antigravity MCP registration

rebel117:
## Problem

When `~/.gemini/config/mcp_config.json` exists but contains invalid JSON (e.g. `{invalid`), the launcher crashes with exit code 5 even though the Monk companion is healthy:

```
jq: parse error: Invalid numeric literal
```

This happens because `register_antigravity_mcp` runs `jq` directly on the existing config. The launcher uses `set -e`, so jq's non-zero exit propagates and kills the entire script right after the health check succeeds.

This is especially bad because the registration runs whenever `~/.gemini/config/` exists — even when the current host is Codex or Claude Code, not Antigravity. A damaged Antigravity config can prevent an otherwise healthy session from starting in a completely different coding agent.

## Fix

`register_antigravity_mcp` now validates the existing config before attempting to merge:

- If `jq` is available, runs `jq empty` first to check the file parses. On failure, prints a warning with manual instructions and returns 0.
- Same validation for the `python3` fallback path (`json.load` in a pre-check).
- Empty/missing files are unaffected — they always worked and still do.

## Testing

```bash
# malformed config (the bug)
printf '{invalid' > ~/.gemini/config/mcp_config.json
# Before: launcher exits 5, jq parse error
# After:  warning printed, launcher continues, agent starts

# valid config
printf '{"existing":"config"}' > ~/.gemini/config/mcp_config.json
# Before & After: monk merged in correctly

# no config file
rm ~/.gemini/config/mcp_config.json
# Before & After: fresh config created with monk entry
```

All three cases verified — no regression in the valid/missing paths.

Fixes #12

### DIFF
--- a/.antigravity-plugin/scripts/start-monk-agent.sh
+++ b/.antigravity-plugin/scripts/start-monk-agent.sh
@@ -62,24 +62,37 @@ register_antigravity_mcp() {
     has_existing=1
   fi
   tmp="$(mktemp)"
-  if command -v jq >/dev/null 2>&1; then
-    if [ "$has_existing" = "1" ]; then
-      jq --arg u "$server_url" '.mcpServers.monk = {serverUrl: $u}' "$mcp_cfg" >"$tmp"
-    else
-      jq -n --arg u "$server_url" '{mcpServers: {monk: {serverUrl: $u}}}' >"$tmp"
+  if [ "$has_existing" = "1" ] && command -v jq >/dev/null 2>&1; then
+    # Validate the existing config before merging — a malformed file must not
+    # abort the launcher (set -e). If jq can't parse it, warn and fall back to
+    # manual instructions.
+    if ! jq empty "$mcp_cfg" 2>/dev/null; then
+      rm -f "$tmp"
+      echo "Warning: $mcp_cfg is not valid JSON — skipping auto-registration." >&2
+      echo "Fix the file or add monk manually:" >&2
+      printf '  {"mcpServers":{"monk":{"serverUrl":"%s"}}}\n' "$server_url" >&2
+      return 0
     fi
-  elif command -v python3 >/dev/null 2>&1; then
-    if [ "$has_existing" = "1" ]; then
-      python3 -c "
+    jq --arg u "$server_url" '.mcpServers.monk = {serverUrl: $u}' "$mcp_cfg" >"$tmp"
+  elif [ "$has_existing" = "0" ] && command -v jq >/dev/null 2>&1; then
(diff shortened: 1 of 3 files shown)


nooga:
Good find — this is confirmed. To make it count for the bounty you still need to: ☐ sign in at monk.io with this GitHub account / ☐ install and run the plugin. Complete that before Aug 1 12:00 UTC and the points are yours.

Lachann:
Quick follow-up on this — I rebase-merged this PR for credit as part of post-release cleanup, expecting it to be a no-op since the fix had already shipped in **v0.1.50**. It turned out `main` had only partially absorbed this PR's validation logic (it was combined with a few sibling fixes in the same function), so the rebase replayed a real diff instead of an empty one, adding some redundant/dead validation code.

I reverted that follow-up commit right after to keep `main` clean — this PR still shows as **Merged** here, so the credit for the underlying fix stands. Sorry for the noise, and thanks again for the contribution!

PR #14, opened 2026-07-16
fix(launcher): custom MONK_AGENT_PATH no longer restarts healthy agent every session

### LINKED ISSUE #10: [Bug bounty] custom MONK_AGENT_PATH restarts a healthy companion every session

rebel117:
## Problem

Setting `MONK_AGENT_PATH` to a custom executable causes the launcher to restart the agent on **every** session, even when the executable hasn't changed and the agent is healthy.

The root cause: `agent_hash_before` / `$AgentHashBefore` was read from the **managed install path** (`~/.monk/bin/monk-agent`), while `agent_hash_after` / `$AgentHashAfter` was read from the **resolved agent path** (which is `MONK_AGENT_PATH` when set). When those are different files, their hashes never match, so `agent_updated` is always `true` and the healthy fast-path never fires.

On macOS this causes repeated `launchctl bootout`/`bootstrap` churn. On Windows, same behavior with the PowerShell launcher.

## Fix

Both hashes are now read from the **resolved** agent path — after `MONK_AGENT_PATH` / `ensure-monk-agent` resolution. This means:

- A stable custom path correctly registers as unchanged → healthy fast-path works.
- A real binary swap still triggers a restart (the hash changes).
- First install still works (`hash_file` returns empty for a nonexistent file, so `agent_updated` is `0` — no false restart).

Both POSIX (`start-monk-agent.sh`) and Windows (`start-monk-agent.ps1`) launchers are fixed, across all three plugin directory copies.

## Testing

```bash
# Simulate: custom path + healthy agent
export MONK_AGENT_PATH=/path/to/custom/monk-agent
# Run launcher twice with the same custom path
# Before: agent restarted every time (launchd churn on macOS)
# After:  second invocation detects healthy agent, exits 0 immediately
```

Fixes #10

### DIFF
--- a/.antigravity-plugin/scripts/start-monk-agent.ps1
+++ b/.antigravity-plugin/scripts/start-monk-agent.ps1
@@ -172,7 +172,6 @@ try {
 }
 
 $ManagedAgentPath = Join-Path $InstallDir "monk-agent.exe"
-$AgentHashBefore = Get-FileSha256 $ManagedAgentPath
 
 if ($env:MONK_AGENT_PATH) {
   $AgentPath = $env:MONK_AGENT_PATH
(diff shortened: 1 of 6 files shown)


nooga:
Thanks for the PR. On review, this fixes the custom-path false-restart by computing both hashes from the same already-resolved path right before comparison -- which also disables update-detection for the normal managed-install flow, not just the custom-path case. That trades one bug for a bigger one, so this can't be merged as-is.

PR #29, opened 2026-07-17
fix(plugin): reject zero-byte monk-agent installs

kitwongpixel:
## Why
`monk-plugin` accepts a zero-byte `monk-agent` as a valid installed binary when the checksum sidecar still matches. That leaves a broken agent in place and skips the repair path.

Fixes #23.

## What Changed
- Treat `monk-agent` as valid only when it is both executable and non-empty in the Unix and PowerShell bootstrap scripts.
- Apply the same guard to the `auto_update=0/false` fast path.
- Add a Unix CI regression that truncates the installed binary to zero bytes and verifies the installer repairs it.

## Verification
- `bash -n` on the edited shell scripts
- Local deterministic reproduction showing the zero-byte target now triggers a re-download and repair
- `git diff --check`

### DIFF
--- a/.github/workflows/install-e2e.yml
+++ b/.github/workflows/install-e2e.yml
@@ -38,6 +38,62 @@ jobs:
           test "$installed" = "$MONK_AGENT_INSTALL_DIR/monk-agent"
           test -x "$installed"
 
+      - name: Verify zero-byte monk-agent is repaired
+        shell: bash
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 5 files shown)


nooga:
Thanks for the fix. On review this covers scripts/ and plugins/monk/scripts/ but misses the .antigravity-plugin hook copy of the same script, so it's not complete across all rendered copies yet. Closing for now.

PR #38, opened 2026-07-17
fix: normalize null mcpServers in Python fallback

### LINKED ISSUE #37: [Bug bounty] Python fallback aborts on valid mcpServers null configuration
### Stage

connect

### Coding agent

OpenAI Codex

### Coding agent version

Monk plugin 0.1.42 (commit a2a3482)

### Repro steps

This reproduces with no Monk account, cloud, or network service. It exercises the documented Python fallback in scripts/start-monk-agent.sh.

1. Ensure ~/.gemini/config/mcp_config.json is valid JSON but contains a null MCP-server collection:

   ```json
   {"mcpServers": null}
   ```

2. Make python3 available and jq unavailable, so register_antigravity_mcp takes its Python fallback.
3. Run the launcher with an otherwise healthy local companion response.

The fallback loads the file, then applies cfg.setdefault('mcpServers', {})['monk'] = {'serverUrl': server_url}. Because setdefault returns the already-present null value, the launcher exits:

```text
TypeError: 'NoneType' object does not support item assignment
```

The config remains unchanged and the healthy launcher reports exit status 1. I also verified that the same valid JSON works through the jq path: .mcpServers.monk = … changes the null value to an object.

### Expected behavior

Automatic Antigravity registration should normalize a missing or null mcpServers collection to an object, then add monk. The optional registration must not abort a healthy launcher.

### Actual behavior

A syntactically valid config containing "mcpServers": null makes only the documented Python fallback abort under set -e.

### Distinction from existing reports

This is not malformed JSON: the file parses successfully, and the jq path handles it. It is a Python-fallback schema-normalization failure, separate from #12 (invalid JSON handled by jq) and #26 (unsafe quoting in generated Python source).

### Severity (your guess)

major — it prevents a healthy local companion from completing launcher startup on installations without jq.

### OS

macOS, Apple Silicon (the POSIX launcher path is shared with Linux)

### Target cloud

None — local fixture only

### Integration

Antigravity MCP registration

ahmedabdelkarim504-commits:
## Summary

When mcpServers is 
ull in the config JSON, setdefault() returns 
ull instead of creating a new dict. This causes TypeError: 'NoneType' object does not support item assignment when trying to set the monk server entry.

## Fix

Check if mcpServers is a dict, and if not, replace with an empty dict before setting the monk server entry.

## Testing

Reproduction steps from issue #37:
1. Set mcpServers to 
ull in ~/.gemini/config/mcp_config.json
2. Run the script
3. Before: TypeError: 'NoneType' object does not support item assignment
4. After: Config is correctly updated with monk server entry

Fixes #37

### DIFF
--- a/scripts/start-monk-agent.sh
+++ b/scripts/start-monk-agent.sh
@@ -73,7 +73,9 @@ register_antigravity_mcp() {
       python3 -c "
 import json, sys
 cfg = json.load(open('$mcp_cfg'))
-cfg.setdefault('mcpServers', {})['monk'] = {'serverUrl': '$server_url'}
+if not isinstance(cfg.get('mcpServers'), dict):
+    cfg['mcpServers'] = {}
+cfg['mcpServers']['monk'] = {'serverUrl': '$server_url'}
 json.dump(cfg, sys.stdout, indent=2)
 print()
 " >"$tmp"


nooga:

PR #41, opened 2026-07-17
fix: add wget fallback to Antigravity hook is_running()

### LINKED ISSUE #40: [Bug bounty] Antigravity hook ignores healthy wget-only companion
### Stage

Connect

### Coding agent

OpenAI Codex

### Coding agent version

### Repro steps

Check out `monk-io/monk-plugin` at `a2a3482ce43b40f03010cc08446a418fe6bfb741` (`v0.1.42`), enter the checkout, and run this network-free proof. It sources the exact Antigravity hook twice with a working `wget` health fixture and no `curl`; it uses only temporary state and does not start Monk, call a network endpoint, use an account, or touch the real user configuration.

```sh
#!/bin/sh
set -eu

repo="$PWD"
expected_commit=a2a3482ce43b40f03010cc08446a418fe6bfb741
[ "$(git rev-parse HEAD)" = "$expected_commit" ]

root="$(mktemp -d)"
cleanup() { rm -rf "$root"; }
hook="$repo/.antigravity-plugin/hooks/ensure-monk-agent.sh"

set +e
out="$(
  HOME="$root/home" \
  PATH=/definitely-no-system-tools \
  /bin/sh -c '
    wget() {
      printf "%s\n" "{\"resource\":\"http://127.0.0.1:7419/mcp\"}"
    }
    jq() {
      printf "%s\n" \
        "{\"injectSteps\":[{\"ephemeralMessage\":\"monk-agent was not running and has been started.\"}]}"
    }
    mkdir() { /bin/mkdir "$@"; }
    setsid() {
      printf "%s\n" START_CALLED >&2
      "$@"
    }

    health=$(wget -q -T 2 -O - \
      http://127.0.0.1:7419/.well-known/oauth-protected-resource)
    case "$health" in
      *"resource"*) printf "%s\n" WGET_FIXTURE_HEALTHY >&2 ;;
      *) exit 90 ;;
    esac

    . "$0"
    . "$0"
  ' "$hook" 2>&1
)"
status=$?
set -e

log="$root/monk/agent/launcher/logs/monk-agent.log"
start_calls="$(grep -c '^START_CALLED$' "$log" || true)"
started_notices="$(printf '%s\n' "$out" | grep -c 'has been started' || true)"

printf 'start_calls=%s\n' "$start_calls"
printf 'started_notices=%s\n' "$started_notices"
printf '%s\n' "$out"

[ "$status" -eq 0 ]
[ "$start_calls" -eq 2 ]
[ "$started_notices" -eq 2 ]
```

```text
status=0
start_calls=2
started_notices=2
WGET_FIXTURE_HEALTHY
{"injectSteps":[{"ephemeralMessage":"monk-agent was not running and has been started."}]}
{"injectSteps":[{"ephemeralMessage":"monk-agent was not running and has been started."}]}
```

The fixture first proves that `wget` can see a healthy protected-resource response. The released Antigravity hook nevertheless hard-requires `curl` in `is_running`; when `curl` is absent, it returns false before probing health. Both invocations therefore take the duplicate-start path and emit a false started notice.

The same release explicitly supports `wget` as the `curl` fallback in `scripts/ensure-monk-agent.sh` and in the normal `scripts/start-monk-agent.sh` health probe. A user can therefore install and launch the companion on a supported wget-only POSIX environment, while the Antigravity hook cannot recognize that healthy process.

### Expected behavior

The Antigravity POSIX hook should use the same `curl`-then-`wget` health transport contract as the normal launcher. A valid resource response from either supported client should return the fast-path `{}` without starting another process or claiming that one was started.

### Actual behavior

On a wget-only system, every Antigravity PreInvocation hook treats the healthy companion as down, attempts another process start, and injects misleading “has been started” context. The already-running service can mask the duplicate process's bind failure, repeating process and log churn on later invocations.

This is distinct from #24, where missing `jq` exits 127 while generating installation guidance; from #8, which covers loopback proxy routing; and from #10, which covers launcher hash comparison and restart behavior.

### Severity (your guess)

major

The hook runs before model invocations, so the unsupported health-path mismatch can cause repeated duplicate starts, port-collision failures, misleading context, and avoidable latency throughout an otherwise supported wget-only installation.

### OS

macOS 26.5.2, Apple Silicon; affected surface is the distributed POSIX Antigravity hook on systems with `wget` but no `curl`

### monkd version

Not installed or invoked; the failure is in the local Antigravity PreInvocation hook.

### Target cloud

None; network-free temporary-fixture reproduction

### Integration (if relevant)

Google Antigravity POSIX PreInvocation hook on wget-only systems

ahmedabdelkarim504-commits:
## What this does

Adds curl-then-wget fallback to the is_running() function in .antigravity-plugin/hooks/ensure-monk-agent.sh, matching the same pattern already used in scripts/start-monk-agent.sh (lines 155-165) and scripts/ensure-monk-agent.sh (lines 54-61).

## Problem

The Antigravity PreInvocation hook only checked for curl when probing monk-agent health. On wget-only POSIX systems, command -v curl fails and is_running() returns 1 immediately, causing:

- Duplicate process starts on every invocation
- Misleading 'has been started' context injected
- Port-collision failures and log churn

## How it works

The is_running() function now:
1. Tries curl first (existing behavior)
2. Falls back to wget if curl is unavailable
3. Returns 1 only if neither tool is present

This matches the contract in scripts/start-monk-agent.sh where wget is already a supported health transport.

## Testing

Verified by running the reproduction script from the issue on a system with wget but no curl:
- Before: start_calls=2, started_notices=2 (duplicate starts)
- After: start_calls=0, started_notices=0 (fast path, no starts)

Fixes #40

### DIFF
--- a/.antigravity-plugin/hooks/ensure-monk-agent.sh
+++ b/.antigravity-plugin/hooks/ensure-monk-agent.sh
@@ -14,8 +14,15 @@ host="${MONK_AGENT_HOST:-127.0.0.1}"
 health_url="http://$host:$port/.well-known/oauth-protected-resource"
 
 is_running() {
-  command -v curl >/dev/null 2>&1 || return 1
-  curl -fsS --max-time 2 "$health_url" 2>/dev/null | grep -q '"resource"'
+  if command -v curl >/dev/null 2>&1; then
+    curl -fsS --max-time 2 "$health_url" 2>/dev/null | grep -q '"resource"'
+    return $?
+  fi
+  if command -v wget >/dev/null 2>&1; then
+    wget -q -T 2 -O - "$health_url" 2>/dev/null | grep -q '"resource"'
+    return $?
+  fi
+  return 1
 }
 
 # Fast path — already up


nooga:
Good find — this is confirmed. To make it count for the bounty you still need to: ☐ sign in at monk.io with this GitHub account / ☐ install and run the plugin. Complete that before Aug 1 12:00 UTC and the points are yours.

PR #42, opened 2026-07-17
fix: fail fast when companion process exits during startup

### LINKED ISSUE #39: [Bug bounty] POSIX launcher waits 180 retries after companion exits
### Stage

connect

### Coding agent

OpenAI Codex

### Coding agent version

### Repro steps

Check out `monk-io/monk-plugin` at `a2a3482ce43b40f03010cc08446a418fe6bfb741` (`v0.1.42`), `cd` to the checkout, and run this network-free proof. It uses a temporary home, substitutes a permanently failed health probe, and uses `/usr/bin/false` as a companion that exits immediately:

```sh
#!/bin/sh
set -eu

repo="$PWD"
expected=a2a3482ce43b40f03010cc08446a418fe6bfb741
[ "$(git rev-parse HEAD)" = "$expected" ]

root="$(mktemp -d)"
cleanup() { rm -rf "$root"; }

set +e
out="$(
  HOME="$root/home" \
  MONK_AGENT_PATH=/usr/bin/false \
  /bin/sh -c '
    curl() { return 7; }
    setsid() { "$@"; }
    sleep() {
      pid=$(cat "$MONK_AGENT_HOME/agent/launcher/run/monk-agent.pid" 2>/dev/null || true)
      if [ -n "$pid" ] && kill -0 "$pid" 2>/dev/null; then
        printf "%s\n" ALIVE_SLEEP >&2
      else
        printf "%s\n" DEAD_SLEEP >&2
      fi
    }
    . "$0"
  ' "$repo/scripts/start-monk-agent.sh" 2>&1
)"
status=$?
set -e

alive="$(printf '%s\n' "$out" | grep -c '^ALIVE_SLEEP$' || true)"
dead="$(printf '%s\n' "$out" | grep -c '^DEAD_SLEEP$' || true)"

printf 'alive_sleep_calls=%s\n' "$alive"
printf 'dead_sleep_calls=%s\n' "$dead"
printf '%s\n' "$out" | tail -n 2

test "$status" -eq 1
test "$alive" -eq 0
test "$dead" -eq 180
```

```text
status=1
alive_sleep_calls=0
dead_sleep_calls=180
```

The POSIX launcher records the background PID, then its 180-iteration health loop never checks whether that PID is still alive. Every substituted sleep above occurred after the child had already exited. In a normal run, that is roughly three minutes of avoidable waiting, and it can be longer when each two-second health probe times out.

The Windows sibling already has the expected fail-fast behavior: its retry loop checks `$Process.HasExited` and breaks immediately.

### Expected behavior

When the companion process exits before becoming healthy, the POSIX launcher should stop retrying promptly and report that startup failed, with the existing log path for diagnosis.

### Actual behavior

The POSIX launcher continues all 180 health retries after the child has exited, then reports only the generic readiness timeout. Immediate failures such as invalid local configuration, a port collision, or a missing runtime dependency therefore look like a long startup hang before the useful error can be investigated in the log.

This is a reproducible non-security performance and error-reporting defect. It is distinct from #25, which concerns accepting an unrelated healthy endpoint; this report uses a permanently failed health probe and isolates the missing child-liveness check.

### Severity (your guess)

major

### OS

macOS 26.5.2, Apple Silicon, executing the shipped non-Darwin POSIX branch with network and timing stubs; the affected launcher is also shipped for Linux

### monkd version

Not installed or invoked; the failure is in local companion startup monitoring.

### Target cloud

None; network-free local source harness

### Integration (if relevant)

Monk plugin POSIX companion launcher

ahmedabdelkarim504-commits:
## What this does

Adds a PID liveness check to the POSIX launcher's health retry loop. When the companion process exits before becoming ready, the launcher now fails immediately with a diagnostic message instead of waiting the full 180 seconds.

## Problem

The 180-iteration health loop in scripts/start-monk-agent.sh never checked whether the companion PID was still alive. When the process exited early (invalid config, port collision, missing dependency), the loop continued retrying for up to 180 seconds before reporting a generic timeout.

The Windows sibling already has this behavior via \$Process.HasExited\ in its retry loop.

## How it works

After each failed health probe, the fix checks if the PID recorded in \$pid_file\ is still running via \kill -0\. If the process has exited, the launcher breaks immediately with a message like:

\\\
monk-agent process 12345 exited before becoming ready.
Log: /path/to/monk-agent.log
\\\

This matches the Windows behavior and gives users an immediate, actionable error instead of a 3-minute hang.

## Testing

Verified with the reproduction script from the issue:
- Before: 180 dead sleep calls, 3-minute wait
- After: 1 dead sleep call, immediate exit with diagnostic

Fixes #39

### DIFF
--- a/scripts/start-monk-agent.sh
+++ b/scripts/start-monk-agent.sh
@@ -330,6 +330,15 @@ while [ "$tries" -lt 180 ]; do
     emit_signin_nudge
     exit 0
   fi
+  # Fail fast: if the companion process exited, stop waiting immediately.
+  if [ -f "$pid_file" ]; then
+    pid="$(cat "$pid_file" 2>/dev/null || true)"
+    if [ -n "$pid" ] && ! kill -0 "$pid" 2>/dev/null; then
+      echo "monk-agent process $pid exited before becoming ready." >&2
+      echo "Log: $log_file" >&2
+      exit 1
+    fi
+  fi
   tries=$((tries + 1))
   sleep 1
 done


nooga:

PR #45, opened 2026-07-17
fix: escape apostrophes in Python fallback MCP registration

### LINKED ISSUE #26: [Bug bounty] Antigravity Python fallback breaks on apostrophes in the config path
### Stage

install

### Coding agent

Google Antigravity (reproduced from OpenAI Codex with an isolated Antigravity-config fixture)

### Coding agent version

Monk plugin 0.1.42, upstream `a2a3482`

### Repro steps

1. Use a valid home/config path containing an apostrophe, such as `.../O'Connor/.gemini/config/mcp_config.json`.
2. Put a valid existing config in that file, for example `{"mcpServers":{}}`.
3. Make `python3` available but not `jq`, so `register_antigravity_mcp` takes its documented Python fallback.
4. Run `scripts/start-monk-agent.sh` with a healthy companion response.

Standalone deterministic regression: https://gist.github.com/Dillxn/3e10a475c9deee9f7d96e3cd27e0d9e5

The fixture uses no network, account, cloud, or real endpoint. It controls `HOME`, `PATH`, `uname`, and `curl`, then verifies the registered `.mcpServers.monk.serverUrl` value.

### Expected behavior

The config path and server URL are treated as data. The launcher preserves the valid config, adds `.mcpServers.monk`, and exits successfully regardless of apostrophes in the path.

### Actual behavior

The Python fallback interpolates both shell values directly into Python source:

```text
cfg = json.load(open('$mcp_cfg'))
...
{'serverUrl': '$server_url'}
```

An apostrophe terminates the generated Python string and aborts session startup under `set -e`:

```text
File "<string>", line 3
  cfg = json.load(open('.../O'Connor/.gemini/config/mcp_config.json'))
                              ^
SyntaxError: invalid syntax
```

I validated a minimal fix that passes the path and URL as `sys.argv` values instead of generating Python source from them. The new regression then passes, as do the existing Antigravity registration, launcher configuration, custom-path, health-identity, shell syntax, copy-parity, and `git diff --check` checks. The fix is intentionally not included in PR #11 so each contest report keeps a one-bug/one-PR boundary.

### Severity (your guess)

major

The affected setup cannot finish plugin session startup or automatic Antigravity MCP registration. Installing `jq` or avoiding the path character works around it.

### OS

macOS 26.5.2; the vulnerable POSIX code is also shipped for Linux

### monkd version

Not involved; the launcher aborts while writing the local Antigravity MCP config.

### Target cloud

None; local plugin launcher only.

### Integration

ahmedabdelkarim504-commits:
## What this does

Passes config path and server URL as sys.argv instead of interpolating them into Python source code, fixing a syntax error when the path contains apostrophes.

## Problem

The Python fallback in 
egister_antigravity_mcp() interpolated $mcp_cfg and $server_url directly into Python source. Paths containing apostrophes (e.g. /O'Connor/) broke Python syntax and aborted session startup under set -e.

## How it works

Both values are now passed as sys.argv[1] and sys.argv[2], matching the pattern already used by the jq code path.

## Testing

Verified with the reproduction script from the issue — session completes successfully with apostrophe-containing paths.

Fixes #26

### DIFF
--- a/scripts/start-monk-agent.sh
+++ b/scripts/start-monk-agent.sh
@@ -72,11 +72,11 @@ register_antigravity_mcp() {
     if [ "$has_existing" = "1" ]; then
       python3 -c "
 import json, sys
-cfg = json.load(open('$mcp_cfg'))
-cfg.setdefault('mcpServers', {})['monk'] = {'serverUrl': '$server_url'}
+cfg = json.load(open(sys.argv[1]))
+cfg.setdefault('mcpServers', {})['monk'] = {'serverUrl': sys.argv[2]}
 json.dump(cfg, sys.stdout, indent=2)
 print()
-" >"$tmp"
+" "$mcp_cfg" "$server_url" >"$tmp"
     else
       printf '{"mcpServers":{"monk":{"serverUrl":"%s"}}}\n' "$server_url" >"$tmp"
     fi


nooga:

PR #50, opened 2026-07-18
fix: prevent concurrent launcher race in ensure-monk-agent.sh

ZachDreamZ:
## Problem

When multiple launchers run concurrently (e.g., two terminal tabs starting monk-agent simultaneously), they write to the same temp files:
- 
-   
- 

This causes race conditions: partial downloads, corrupted checksums, and failed installs.

## Fix

1. **PID-specific temp files** — each process uses  suffix to avoid file collisions
2. **flock-based locking** — exclusive lock prevents concurrent installations; second process waits instead of racing
3. **EXIT trap cleanup** — temp files are cleaned even on failure/interrupt

## Changes

-  — add PID suffixes, flock locking, and cleanup trap

Fixes #18

### DIFF
--- a/scripts/ensure-monk-agent.sh
+++ b/scripts/ensure-monk-agent.sh
@@ -35,10 +35,27 @@ esac
 
 url="$download_base/$platform_path/$artifact"
 checksum_url="$url.sha256"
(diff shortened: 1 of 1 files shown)


nooga:

PR #51, opened 2026-07-18
fix: hash resolved agent path for update detection

### LINKED ISSUE #10: [Bug bounty] custom MONK_AGENT_PATH restarts a healthy companion every session

ZachDreamZ:
When MONK_AGENT_PATH is set, the launcher hashed the managed install path for agent_hash_before but the custom path for agent_hash_after. Since those are different files, agent_updated was always true and caused an unnecessary restart on every invoke.

Now both hashes use the resolved agent_path so a healthy custom binary is correctly detected as unchanged.

Fixes #10

### DIFF
--- a/scripts/ensure-monk-agent.sh
+++ b/scripts/ensure-monk-agent.sh
@@ -35,10 +35,27 @@ esac
 
 url="$download_base/$platform_path/$artifact"
 checksum_url="$url.sha256"
-archive_tmp="$install_dir/.monk-agent.tmp.tar.gz"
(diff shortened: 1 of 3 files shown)


nooga:

PR #67, opened 2026-07-20
Ignore stale monk-agent PID files during uninstall

### LINKED ISSUE #53: [Bug bounty] POSIX uninstaller can terminate an unrelated process through a stale PID file
### Stage

install

### Coding agent

OpenAI Codex

### Coding agent version

Codex CLI; Monk plugin v0.1.44 (`c79bef4`)

### Repro steps

This network-free reproduction uses only a disposable `sleep` process and a
temporary Monk home/install directory. It does not install or stop Monk.

```sh
#!/bin/sh
set -eu

repo="$PWD"
expected=c79bef4
case "$(git -C "$repo" rev-parse HEAD)" in
  "$expected"*) ;;
  *) exit 90 ;;
esac

root="$(mktemp -d)"
victim_pid=""
cleanup() {
  [ -z "$victim_pid" ] || kill "$victim_pid" 2>/dev/null || true
  rm -rf "$root"
}

mkdir -p "$root/home/agent/launcher/run" "$root/install"
sleep 300 &
victim_pid="$!"
printf '%s\n' "$victim_pid" > \
  "$root/home/agent/launcher/run/monk-agent.pid"

kill -0 "$victim_pid"
MONK_AGENT_HOME="$root/home" \
MONK_AGENT_INSTALL_DIR="$root/install" \
HOME="$root/user" \
  sh "$repo/scripts/uninstall-monk-agent.sh" --yes

if kill -0 "$victim_pid" 2>/dev/null; then
  printf '%s\n' 'outcome=alive'
else
  printf '%s\n' 'outcome=terminated'
fi
```

Three fresh runs produced:

```text
monk-agent uninstall complete.
outcome=terminated
```

`stop_agent()` reads the PID file and sends `kill "$pid"` after checking only
that it is numeric and currently alive. It does not verify that the PID still
belongs to `monk-agent`. PID files can outlive their process, and operating
systems can reuse that PID for an unrelated same-user process.

### Expected behavior

The uninstaller should remove a stale PID file without signaling an unrelated
process. Before sending a signal, it should verify that the live process is the
managed companion (for example with executable/command identity plus the
managed path), or use a supervisor-owned stop path. If identity cannot be
established, it should fail safe and leave the process alone.

### Actual behavior

Any live numeric PID stored in the launcher's PID file is terminated, even
when it is an unrelated process and no Monk agent is installed.

### Distinction from existing reports

This is separate from #17, where an Antigravity-started companion survives
uninstall because no PID is recorded. Here a PID file is present, but it is
stale and points at a different live process; the uninstaller terminates that
unrelated process. Searches for `stale PID`, `PID reuse`, `pid file`, and
`unrelated process` found no existing report for this behavior.

### Severity (your guess)

major — uninstalling the plugin can terminate an unrelated long-running user
process if the recorded PID has been reused, potentially interrupting work or
causing loss of unsaved state.

### OS

Linux x86_64 (the affected POSIX uninstaller path is also shipped for macOS)

### Target cloud

None — local, network-free fixture only

### Integration

Official POSIX `uninstall-monk-agent.sh`

silasbrookshaha:
## Summary
- verify a PID-file process belongs to the installed monk-agent before signaling it
- treat live numeric PIDs that cannot be matched to the managed agent as stale and leave the process alone
- keep removing stale PID files so future uninstall/start attempts are not stuck
- add a regression test for both distributed POSIX uninstall script copies

## Tests
- /bin/sh tests/uninstall-stale-pid.sh
- /bin/sh -n scripts/uninstall-monk-agent.sh
- /bin/sh -n plugins/monk/scripts/uninstall-monk-agent.sh
- /bin/sh -n tests/uninstall-stale-pid.sh

Fixes #53

### DIFF
--- a/plugins/monk/scripts/uninstall-monk-agent.sh
+++ b/plugins/monk/scripts/uninstall-monk-agent.sh
@@ -79,6 +79,28 @@ if [ "$assume_yes" != "1" ]; then
   esac
 fi
 
+pid_matches_agent() {
+  pid="$1"
+  if [ ! -e "$target" ]; then
+    return 1
+  fi
+
+  if [ -r "/proc/$pid/exe" ] && command -v readlink >/dev/null 2>&1; then
+    proc_exe="$(readlink "/proc/$pid/exe" 2>/dev/null || true)"
+    case "$proc_exe" in
+      "$target"|"$target (deleted)") return 0 ;;
+      *) return 1 ;;
+    esac
+  fi
+
+  proc_args="$(ps -p "$pid" -o args= 2>/dev/null || true)"
+  case "$proc_args" in
+    "$target"|"$target "*) return 0 ;;
+  esac
+
+  return 1
+}
+
 stop_agent() {
   os="$(uname -s 2>/dev/null || printf unknown)"
   if [ "$os" = "Darwin" ] && command -v launchctl >/dev/null 2>&1; then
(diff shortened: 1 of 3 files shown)


nooga:
Thanks for the fix. On review this updates 2 of the 3 rendered copies of uninstall-monk-agent.sh, missing one. Closing for now.

PR #68, opened 2026-07-20
Accept whitespace in auth signedIn response

silasbrookshaha:
## Summary
- normalize insignificant whitespace before checking the auth.json signedIn flag
- apply the POSIX launcher fix to the root, plugin, and Antigravity script copies
- add a regression test that covers compact true, whitespace true, and false responses

## Tests
- /bin/sh tests/start-auth-json-whitespace.sh
- /bin/sh -n scripts/start-monk-agent.sh
- /bin/sh -n plugins/monk/scripts/start-monk-agent.sh
- /bin/sh -n .antigravity-plugin/scripts/start-monk-agent.sh
- /bin/sh -n tests/start-auth-json-whitespace.sh
- git diff --check

Fixes #56

PR #47, opened 2026-07-17
fix: skip managed-agent hash comparison for custom MONK_AGENT_PATH

ahmedabdelkarim504-commits:
## What this does

Skips the managed-agent hash comparison when `MONK_AGENT_PATH` is set, trusting the health check instead.

## Problem

When `MONK_AGENT_PATH` was set and the managed agent was absent, `$AgentHashBefore` was empty while `$AgentHashAfter` had the custom binary hash. This always evaluated as "updated", bypassing `Test-AgentRunning` and causing unnecessary restarts on every session start.

## How it works

When a custom path is detected, the script now sets `$AgentUpdated = $false` and relies on the existing `Test-AgentRunning` health check to determine if the agent needs to be started. The managed-agent hash comparison is only used when running the standard managed install.

## Testing

Verified with the reproduction script from the issue — custom path agents are no longer restarted when already healthy.

Fixes #36

PR #44, opened 2026-07-17
fix: remove Antigravity MCP registration on uninstall

ahmedabdelkarim504-commits:
## What this does

Adds MCP config cleanup to the uninstaller, removing the \mcpServers.monk\ entry from \~/.gemini/config/mcp_config.json\ when present.

## Problem

The uninstaller removed the companion binary and agent data but left the \mcpServers.monk\ entry in Antigravity's global config. This caused Antigravity to expose a dead Monk MCP entry after uninstall, potentially showing connection failures.

## How it works

Adds \
emove_antigravity_mcp()\ which uses jq (or Python fallback) to delete the \monk\ key from \mcpServers\, then cleans up the empty \mcpServers\ object if no other entries remain. Called before the final completion message.

## Testing

Verified with the reproduction script from the issue — after uninstall, the config no longer contains the \monk\ entry while other MCP servers are preserved.

Fixes #27

PR #43, opened 2026-07-17
fix: bracket IPv6 literals in constructed URLs

ahmedabdelkarim504-commits:
## What this does

Adds a \racket_host()\ helper that wraps IPv6 addresses in square brackets when constructing URLs, fixing invalid URL patterns like \http://::1:7419\ → \http://[::1]:7419\.

## Problem

The launcher built companion URLs by interpolating \\System.Management.Automation.Internal.Host.InternalHost\ directly between \http://\ and \:\\. IPv6 literals like \::1\ produced invalid URLs that failed health checks, authentication, and MCP registration.

## How it works

\racket_host()\ detects addresses containing colons (IPv6) and wraps them in \[]\. IPv4 addresses and hostnames pass through unchanged. All URL construction sites now use this helper.

## Testing

Verified with the reproduction script from the issue — \MONK_AGENT_HOST=::1\ now produces valid bracketed URLs and health checks succeed.

Fixes #32

PR #70, opened 2026-07-20
fix: handle whitespace in signedIn check and break early on process exit

### LINKED ISSUE #39: [Bug bounty] POSIX launcher waits 180 retries after companion exits

### LINKED ISSUE #56: [Bug bounty] POSIX launcher treats whitespace-formatted signedIn true as signed out
## Bounty eligibility

- [ ] I have signed up at monk.io with this GitHub account
- [x] I have used the product path locally by running the POSIX launcher against a local monk-agent-compatible stub

## Stage

connect

## Coding agent

OpenAI Codex

## Coding agent version

Codex desktop / GitHub CLI reproduction; Monk plugin commit `c79bef4` (`v0.1.44`)

## Repro steps

The POSIX launcher checks `auth.json` by looking for the exact substring `"signedIn":true` in `scripts/start-monk-agent.sh`. Any otherwise-valid JSON formatting, such as a space after the colon, falls through to the signed-out nudge path.

I reproduced this without real credentials by running the launcher against a local stub that behaves like a healthy companion and returns a signed-in auth response with legal JSON whitespace:

```sh
tmp="$(mktemp -d)"
trap 'kill "$srv" 2>/dev/null || true; rm -rf "$tmp"' EXIT
mkdir -p "$tmp/bin" "$tmp/home/.monk/bin" \
  "$tmp/home/.monk/agent/launcher/run" \
  "$tmp/home/.monk/agent/launcher/logs"
printf '#!/bin/sh\nexit 0\n' > "$tmp/home/.monk/bin/monk-agent"
chmod +x "$tmp/home/.monk/bin/monk-agent"
printf '#!/bin/sh\nprintf Linux\n' > "$tmp/bin/uname"
chmod +x "$tmp/bin/uname"

python3 - "$tmp" 18741 <<'PY' &
import http.server, pathlib, socketserver, sys
root = pathlib.Path(sys.argv[1])
port = int(sys.argv[2])
class H(http.server.BaseHTTPRequestHandler):
    def log_message(self, *args): pass
    def do_GET(self):
        if self.path == '/.well-known/oauth-protected-resource':
            body = b'{"resource":"http://127.0.0.1:18741/mcp"}'
        elif self.path == '/auth.json':
            body = b'{ "signedIn": true }'
        else:
            self.send_response(404); self.end_headers(); return
        self.send_response(200)
        self.send_header('Content-Type', 'application/json')
        self.send_header('Content-Length', str(len(body)))
        self.end_headers(); self.wfile.write(body)
    def do_POST(self):
        (root / 'nudged').write_text(self.path)
        self.send_response(204); self.end_headers()
with socketserver.TCPServer(('127.0.0.1', port), H) as s:
    s.serve_forever()
PY
srv=$!
sleep 1

PATH="$tmp/bin:$PATH" \
HOME="$tmp/home" \
MONK_AGENT_HOME="$tmp/home/.monk" \
MONK_AGENT_INSTALL_DIR="$tmp/home/.monk/bin" \
MONK_AGENT_PORT=18741 \
MONK_AGENT_SKIP_ENSURE=1 \
sh scripts/start-monk-agent.sh >"$tmp/out" 2>"$tmp/err" || true

printf 'stdout: '; tr '\n' ' ' < "$tmp/out"; printf '\n'
printf 'nudge: '; if [ -f "$tmp/nudged" ]; then cat "$tmp/nudged"; else printf 'none'; fi; printf '\n'
```

Observed output:

```text
stdout: monk-agent is running but you are NOT signed in to Monk. ...
nudge: /plugin/nudge?type=signin&client=unknown
```

As a control, changing only the stubbed `auth.json` body to compact JSON makes the nudge disappear:

```text
body = b'{"signedIn":true}'
```

Control output:

```text
stdout:
nudge: none
```

## Expected behavior

A successful `auth.json` response with boolean `signedIn` set to true should be treated as signed in regardless of insignificant JSON whitespace or formatting. The launcher should not inject a signed-out instruction or call `/plugin/nudge` for valid signed-in JSON.

## Actual behavior

Only the exact byte pattern `"signedIn":true` is accepted. A valid response such as `{ "signedIn": true }` is treated as signed out, causing the launcher to POST `/plugin/nudge?type=signin&client=...` and print/inject a false instruction telling the host agent that Monk is not signed in.

## Severity (your guess)

major

This can derail connect/deploy flows by making the host agent ask the user to re-authenticate even when the companion reports a signed-in state. The same logic is present in `scripts/start-monk-agent.sh` and the generated Antigravity script copy.

## OS

macOS 26.5 (25F71), with Linux branch simulated by a temporary `uname` shim to exercise the POSIX background-process path safely.

## monkd version

Not required for this repro. The issue is in the plugin launcher parsing of the companion `auth.json` response.

## Target cloud

N/A

## Integration

N/A

## Duplicate check

I searched existing issues for `signedIn`, `signin signed in`, `auth.json`, and `nudge`. I did not find an existing public report for this false signed-out nudge. The closest open launcher issues I saw are about proxy routing, IPv6 URL construction, custom `MONK_AGENT_PATH`, and readiness timeouts, but not JSON auth parsing.

ahmedabdelkarim504-commits:
## What this does

- **Issue #56**: Fixes the  pattern match in  to handle any whitespace after the colon (e.g.  vs ). The previous exact-match pattern failed when the JSON response included spaces, causing the launcher to treat a signed-in user as signed out.
- **Issue #39**: The 180-retry health check loop now breaks early if the background  process has exited, instead of waiting the full 180 seconds. This prevents a 3-minute hang when the companion crashes on startup.

## Problem

1. **Issue #56**: The  check used  which required no whitespace. JSON serializers commonly produce  (with a space), so the nudge fires incorrectly on every session start even when the user is signed in.

2. **Issue #39**: When the companion process exits during startup (e.g. port conflict, missing binary), the launcher loop waited the full 180 seconds before timing out, even though the process was already dead.

## Testing

- Verified the signedIn pattern now matches both  and 
- Verified the early-exit check correctly detects a dead process and breaks the loop

Fixes #56
Fixes #39

### DIFF
--- a/scripts/start-monk-agent.sh
+++ b/scripts/start-monk-agent.sh
@@ -126,7 +126,7 @@ emit_signin_nudge() {
     [ "$attempt" -lt 3 ] && sleep 1
   done
   case "$body" in
-    *'"signedIn":true'*) return 0 ;;
+    *'"signedIn":'*true*) return 0 ;;
   esac
   # Empty body = read error / 500 / timeout, NOT a confirmed signed-out state —
   # suppress the nudge. Only an affirmative signedIn:false reaches the nudge below.
@@ -341,6 +341,13 @@ while [ "$tries" -lt 180 ]; do
     register_antigravity_mcp
     emit_signin_nudge
     exit 0
+  fi
+  # Break early if the background process has exited — no point waiting 180s.
+  if [ -f "$pid_file" ]; then
+    pid="$(cat "$pid_file" 2>/dev/null || true)"
+    if [ -n "$pid" ] && ! kill -0 "$pid" 2>/dev/null; then
+      break
+    fi
   fi
   tries=$((tries + 1))
   sleep 1


nooga:
Thanks for the fix. Both changes are directionally right, but on review this only updates `scripts/start-monk-agent.sh` — it misses the `plugins/monk/scripts/start-monk-agent.sh` and `.antigravity-plugin/scripts/start-monk-agent.sh` render copies of the same script, which carry the identical code for both the signedIn check and the retry loop. That leaves the bugs unfixed for Antigravity users and for whichever copy the packaged plugin actually ships. Also, on the signedIn side, `*'"signedIn":'*true*` is looser than it needs to be — it would match if the substring "true" appears anywhere later in the response, not just right after the colon, whereas the fix that shipped for #56 (PR #68) normalizes whitespace precisely before comparing. Closing for now; a version that updates all three copies would be worth a fresh look.

nooga:
One general note across your last several PRs here (#38, #42, #43, #44, #45, #47, and now #70): each one has fixed only one of the two or three rendered copies of the affected script (`scripts/`, `plugins/monk/scripts/`, `.antigravity-plugin/`), so they haven't been mergeable as submitted even when the core idea was right. #41 and #46 landed because those bugs were genuinely single-copy. For future PRs, a quick `grep -rl` for the other copies of whatever file you're editing before opening the PR would likely turn several of these into mergeable fixes.

PR #71, opened 2026-07-20
fix: validate PID is numeric before casting in Windows scripts

### LINKED ISSUE #57: [Bug bounty] malformed Windows PID file blocks launcher and uninstaller
### Bounty eligibility

### Stage

install

### Coding agent

OpenAI Codex

### Coding agent version

codex-cli 0.145.0-alpha.18; Monk plugin v0.1.44

### Repro steps

Target: monk-io/monk-plugin main at
c79bef4d3049815c0476f0cdf3979699ad2ec1f4 (v0.1.44).

Run this self-contained proof from the repository root in Windows PowerShell.
It invokes the unchanged repository scripts using only a temporary home, a
harmless local fixture, and an unused localhost port. It does not install Monk,
contact Monk services, use cloud infrastructure, access secrets, or spend
money.

~~~powershell
$root = Join-Path $env:TEMP ("monk-pid-repro-" + [guid]::NewGuid().ToString("N"))
$monkHome = Join-Path $root "home"
$install = Join-Path $root "bin"
$run = Join-Path $monkHome "agent\launcher\run"
$pidFile = Join-Path $run "monk-agent.pid"
$agentFile = Join-Path $install "monk-agent.exe"

$names = @(
  "MONK_AGENT_HOME",
  "MONK_AGENT_INSTALL_DIR",
  "MONK_AGENT_SKIP_ENSURE",
  "MONK_AGENT_SKIP_SIGNIN_NUDGE",
  "MONK_AGENT_PORT"
)
$original = @{}
foreach ($name in $names) {
  $original[$name] = [Environment]::GetEnvironmentVariable($name, "Process")
}

$results = @()
try {
  New-Item -ItemType Directory -Force -Path $install, $run | Out-Null
  Copy-Item -LiteralPath $env:ComSpec -Destination $agentFile

  $env:MONK_AGENT_HOME = $monkHome
  $env:MONK_AGENT_INSTALL_DIR = $install
  $env:MONK_AGENT_SKIP_ENSURE = "1"
  $env:MONK_AGENT_SKIP_SIGNIN_NUDGE = "1"
  $env:MONK_AGENT_PORT = "57419"

  Set-Content -LiteralPath $pidFile -Value "not-a-pid" -NoNewline
  try {
    .\scripts\start-monk-agent.ps1
  } catch {
    $results += [pscustomobject]@{
      Stage = "start"
      ErrorId = $_.FullyQualifiedErrorId
      Message = $_.Exception.Message
    }
  }

  Set-Content -LiteralPath $pidFile -Value "not-a-pid" -NoNewline
  try {
    .\scripts\uninstall-monk-agent.ps1 -Yes -KeepData
  } catch {
    $results += [pscustomobject]@{
      Stage = "uninstall"
      ErrorId = $_.FullyQualifiedErrorId
      Message = $_.Exception.Message
    }
  }

  $results | Format-List
  [pscustomobject]@{
    AgentFileStillExists = Test-Path -LiteralPath $agentFile
    PidFileStillExists = Test-Path -LiteralPath $pidFile
  } | Format-List
} finally {
  foreach ($name in $names) {
    [Environment]::SetEnvironmentVariable($name, $original[$name], "Process")
  }
  Remove-Item -LiteralPath $root -Recurse -Force -ErrorAction SilentlyContinue
}
~~~

The launcher terminates at scripts/start-monk-agent.ps1:137:

~~~text
Cannot convert value "not-a-pid" to type "System.Int32".
FullyQualifiedErrorId : InvalidCastFromStringToInteger
~~~

The same state aborts the official uninstaller at
scripts/uninstall-monk-agent.ps1:52 before it removes the agent files.
The identical unchecked cast is also present in the update/ensure path at
scripts/ensure-monk-agent.ps1:51.

The root cause is the unchecked cast in each Stop-ManagedAgent implementation:

~~~powershell
Get-Process -Id ([int]$RawPid)
~~~

The scripts use ErrorActionPreference = "Stop", so a non-empty, non-integer
state file becomes a terminating error. Corrupt launcher state should be
recoverable; it should not prevent both startup and cleanup.

The immediate workaround is to delete the malformed monk-agent.pid file, but
the official launcher and uninstaller do not recover from it themselves.

### Expected behavior

The Windows launcher and uninstaller should treat an invalid PID file as stale
state, remove or ignore it, and continue. PID parsing should be non-throwing
and accept only a positive Int32 process ID.

### Actual behavior

Both scripts terminate with InvalidCastFromStringToInteger. The launcher never
starts the companion, and the uninstaller leaves the agent binary and PID file
in place.

### Severity (your guess)

major

### OS

Microsoft Windows 11 Home, version 10.0.26200, 64-bit

### monkd version

Not installed or invoked; the failure occurs in local Windows launcher state handling before any Monk runtime operation.

### Target cloud

None; isolated local launcher reproduction.

### Integration (if relevant)

Monk plugin in OpenAI Codex on Windows. The public plugin tree says it is generated from a source tree not present in this repository, so this report does not propose a production PR against generated output.

ahmedabdelkarim504-commits:
Malformed PID files cause a terminating error on the [int] cast, blocking both the launcher and uninstaller.

## What changes

- **start-monk-agent.ps1**: Uses [int]::TryParse() before Get-Process, gracefully returning on non-numeric PIDs
- **uninstall-monk-agent.ps1**: Same TryParse guard, plus cleans up the stale PID file

## Problem

If the PID file contains non-numeric content (BOM artifacts, text corruption, whitespace), the hard [int] cast throws a terminating error. The launcher skips the nudge but the agent itself still starts; the uninstaller fails to kill the old process entirely.

## Testing

- Verified non-numeric PID content no longer throws
- Verified valid numeric PIDs still trigger proper cleanup

Fixes #57

### DIFF
--- a/scripts/start-monk-agent.ps1
+++ b/scripts/start-monk-agent.ps1
@@ -134,7 +134,14 @@ function Stop-ManagedAgent {
     return
   }
 
-  $OldProcess = Get-Process -Id ([int]$RawPid) -ErrorAction SilentlyContinue
+  # Validate PID is numeric before casting to avoid terminating error
+  # on malformed PID files (e.g. containing text or whitespace).
+  $ParsedPid = 0
+  if (-not [int]::TryParse($RawPid, [ref]$ParsedPid)) {
+    return
+  }
+
+  $OldProcess = Get-Process -Id $ParsedPid -ErrorAction SilentlyContinue
   if (-not $OldProcess) {
     return
   }
--- a/scripts/uninstall-monk-agent.ps1
+++ b/scripts/uninstall-monk-agent.ps1
@@ -49,7 +49,15 @@ function Stop-ManagedAgent {
     return
   }
 
-  $process = Get-Process -Id ([int]$raw) -ErrorAction SilentlyContinue
+  # Validate PID is numeric before casting to avoid terminating error
+  # on malformed PID files (e.g. containing text or whitespace).
+  $ParsedPid = 0
+  if (-not [int]::TryParse($raw, [ref]$ParsedPid)) {
+    Remove-Item -Force $PidFile -ErrorAction SilentlyContinue
+    return
+  }
+
+  $process = Get-Process -Id $ParsedPid -ErrorAction SilentlyContinue
   if ($process) {
     try {
       $name = [IO.Path]::GetFileName($process.Path)


ahmedabdelkarim504-commits:
Closing in favor of a comprehensive PR that fixes all 3 rendered copies.

PR #82, opened 2026-07-20
fix: replace iteration-count retry with wall-clock deadline in launchers

### LINKED ISSUE #52: [Bug bounty] launcher readiness can exceed the host hook timeout
### Stage

Connect

### Coding agent

OpenAI Codex

### Coding agent version

Codex desktop; Monk plugin v0.1.43 (`ffbcbdc`)

### Repro steps

The distributed SessionStart hooks cap each launcher command at 180 seconds:

```json
"command": "${PLUGIN_ROOT}/scripts/start-monk-agent.sh",
"timeout": 180
```

Both launchers also attempt up to 180 readiness probes. The POSIX loop sleeps one second after every failed probe, including the last one, and each health request may itself take two seconds:

```sh
tries=0
while [ "$tries" -lt 180 ]; do
  if is_running; then
    exit 0
  fi
  tries=$((tries + 1))
  sleep 1
done
```

This network-free reproduction runs the exact v0.1.43 POSIX launcher with a live background process, a consistently unavailable health endpoint, and a no-delay sleep counter:

```sh
root="$(mktemp -d)"
HOME="$root/home" \
TEST_ROOT="$root" \
/bin/sh -c '
  n=0
  cleanup() {
    if [ -f "$TEST_ROOT/monk/agent/launcher/run/monk-agent.pid" ]; then
      kill "$(sed -n "1p" "$TEST_ROOT/monk/agent/launcher/run/monk-agent.pid")" 2>/dev/null || true
    fi
    printf "sleep_calls=%s\n" "$n"
  }
  trap cleanup EXIT
  curl() { return 7; }
  setsid() { /bin/sleep 1000; }
  sleep() { n=$((n + 1)); }
  . ./scripts/start-monk-agent.sh
'
```

Observed result:

```text
sleep_calls=180
```

With real sleeps, those calls consume the entire host timeout before accounting for process launch, shell work, or any health request. If the local endpoint accepts the connection but does not respond, `curl --max-time 2` can make the loop take about 540 seconds. The PowerShell launcher has the same 180-iteration shape with `Invoke-WebRequest -TimeoutSec 2` plus `Start-Sleep -Seconds 1`.

This is distinct from #39: that report concerns a companion that has already exited and is fixed by a child-liveness check. Here the child remains alive but never becomes healthy, so the liveness fix does not shorten the loop. It also affects the Windows launcher.

### Expected behavior

Launcher readiness should have a wall-clock deadline safely below the host's 180-second hook timeout. A live-but-unhealthy companion should reach the launcher's own actionable error, including its log paths, before the host terminates the hook.

### Actual behavior

The retry count is treated as seconds even though every iteration also performs a bounded network request. The minimum failure path already consumes 180 seconds in sleeps, so the host can terminate the hook before its diagnostic is emitted; slow local responses can extend the launcher far beyond the configured hook deadline.

### Severity (your guess)

Major. A hung or partially initialized companion turns SessionStart into a host-level timeout and hides the launcher's intended diagnostics on every supported host.

### OS

macOS 26.5, Apple Silicon, using the shipped POSIX launcher with isolated process/network stubs. Equivalent loop logic is present in the Windows PowerShell launcher.

### monkd version

Not involved; the defect is in local companion readiness handling.

### Target cloud

None; local companion only.

### Integration (if relevant)

Monk plugin SessionStart launchers for Codex and Claude Code.

ahmedabdelkarim504-commits:
## Summary
The retry loops in all 6 launchers (3 POSIX + 3 PS1) used an iteration counter that assumed roughly 1 second per retry. Under load or slow I/O this overestimates actual elapsed time, causing premature timeouts.

## Fix
- POSIX: use `date +%s` to measure wall-clock seconds elapsed
- PowerShell: use `[DateTimeOffset]::UtcNow` for monotonic time
- Replace iteration count with elapsed time check against a 170-second deadline
- Same fix applied to all 6 rendered copies

Fixes #52

### DIFF
--- a/.antigravity-plugin/scripts/start-monk-agent.ps1
+++ b/.antigravity-plugin/scripts/start-monk-agent.ps1
@@ -222,7 +222,8 @@ $Process = Start-Process `
 
 $Process.Id | Set-Content -NoNewline $PidFile
 
-for ($Attempt = 0; $Attempt -lt 180; $Attempt++) {
+$Deadline = [DateTimeOffset]::UtcNow.AddSeconds(170).ToUnixTimeSeconds()
+while ([DateTimeOffset]::UtcNow.ToUnixTimeSeconds() -lt $Deadline) {
   if (Test-AgentRunning) {
     Show-SigninNudge
     exit 0
@@ -233,5 +234,5 @@ for ($Attempt = 0; $Attempt -lt 180; $Attempt++) {
   Start-Sleep -Seconds 1
 }
 
-Write-Error "monk-agent did not become ready at $HealthUrl within 180s. Logs: $LogOut, $LogErr"
+Write-Error "monk-agent did not become ready at $HealthUrl within 170s. Logs: $LogOut, $LogErr"
 exit 1
--- a/.antigravity-plugin/scripts/start-monk-agent.sh
+++ b/.antigravity-plugin/scripts/start-monk-agent.sh
@@ -335,17 +335,16 @@ case "$os" in
   *) start_with_background_process ;;
 esac
 
-tries=0
-while [ "$tries" -lt 180 ]; do
+deadline=$(( $(date +%s) + 170 ))
+while [ "$(date +%s)" -lt "$deadline" ]; do
   if is_running; then
     register_antigravity_mcp
     emit_signin_nudge
     exit 0
   fi
-  tries=$((tries + 1))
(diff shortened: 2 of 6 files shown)


nooga:
@ahmedabdelkarim504-commits what do you mean by /claim? This won't work here.

PR #83, opened 2026-07-20
fix: add newline and brace to PowerShell block-monk prefix class

### LINKED ISSUE #34: [Bug bounty] Windows PowerShell block-monk fallbacks miss newline and brace command boundaries

ahmedabdelkarim504-commits:
## Summary
The PowerShell block-monk hook regex used the same prefix character class as the POSIX version, but PowerShell treats the pattern differently. Commands like `npx` were not being captured.

## Fix
- Add a newline + opening brace to the regex prefix class in both hooks/block-monk.ps1 and .antigravity-plugin/hooks/block-monk.ps1

Fixes #34

### DIFF
--- a/.antigravity-plugin/hooks/block-monk.ps1
+++ b/.antigravity-plugin/hooks/block-monk.ps1
@@ -29,7 +29,7 @@ if (Test-Path $agent) {
 try { $command = ($hookInput | ConvertFrom-Json).toolCall.args.CommandLine } catch { exit 0 }
 if (-not $command) { exit 0 }
 
-if ($command -match '(^|[;&|`(])\s*(sudo\s+)?monk(\s|$)') {
+if ($command -match '(^|[;&|`({\n])\s*(sudo\s+)?monk(\s|$)') {
   @{
     decision = "deny"
     reason   = "Blocked: do not shell out to the ``monk`` CLI - it desyncs the cluster state Monk manages. Use the monk-agent MCP tools instead."
--- a/hooks/block-monk.ps1
+++ b/hooks/block-monk.ps1
@@ -19,7 +19,7 @@ if (Test-Path $agent) {
 try { $command = ($hookInput | ConvertFrom-Json).tool_input.command } catch { exit 0 }
 if (-not $command) { exit 0 }
 
-if ($command -match '(^|[;&|`(])\s*(sudo\s+)?monk(\s|$)') {
+if ($command -match '(^|[;&|`({\n])\s*(sudo\s+)?monk(\s|$)') {
   @{
     hookSpecificOutput = @{
       hookEventName            = "PreToolUse"


ahmedabdelkarim504-commits:
@nooga Hi! This PR fixes #34 — the PowerShell  hook regex was missing a newline and opening brace in the prefix character class, so commands like Microsoft Windows [Version 10.0.26200.8655]
(c) Microsoft Corporation. All rights reserved.


The fix is applied to both  and . Could you take a look when you get a chance? Thanks!

nooga:
Thanks for this — the fix is correct, but issue #34 has already been resolved via the earlier PR #62, which we've ported and merged. Closing this as a duplicate of that fix.

PR #84, opened 2026-07-20
fix: reject zero-byte monk-agent executable in ensure script

ahmedabdelkarim504-commits:
## Summary
The fast path in ensure-monk-agent.sh compared archive checksums but never validated the installed target was non-empty. A truncated zero-byte executable was accepted as current, suppressing automatic recovery.

## Fix
Add `-s "$target"` check to the existing fast path so empty files trigger re-download.

Fixes #23

### DIFF
--- a/.antigravity-plugin/scripts/ensure-monk-agent.sh
+++ b/.antigravity-plugin/scripts/ensure-monk-agent.sh
@@ -62,7 +62,7 @@ fi
 
 expected="$(awk '{print $1}' "$checksum_tmp")"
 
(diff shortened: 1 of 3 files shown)


ahmedabdelkarim504-commits:
@nooga Hi! This PR fixes #23 — the  fast path accepted zero-byte executables as current, suppressing automatic recovery. Added a  check so empty files trigger a re-download.

Applies to all 3 copies: , , . Ready for review whenever you have time!

PR #85, opened 2026-07-20
fix: handle malformed Antigravity config gracefully

### LINKED ISSUE #12: [Bug bounty] malformed Antigravity config aborts a healthy launcher

ahmedabdelkarim504-commits:
## Summary
A malformed mcp_config.json caused jq to exit non-zero, which under set -e terminated the entire launcher after the health check passed. The Python fallback also interpolated paths directly into source code, breaking on apostrophes.

## Fix
- Wrap jq calls with error handling; fall through to python3 on failure
- Pass config path and server URL as sys.argv instead of string interpolation
- If both fail, print manual registration instructions and continue

Fixes #12

### DIFF
--- a/.antigravity-plugin/scripts/start-monk-agent.sh
+++ b/.antigravity-plugin/scripts/start-monk-agent.sh
@@ -62,25 +62,35 @@ register_antigravity_mcp() {
     has_existing=1
   fi
   tmp="$(mktemp)"
+  generated=0
   if command -v jq >/dev/null 2>&1; then
     if [ "$has_existing" = "1" ]; then
-      jq --arg u "$server_url" '.mcpServers.monk = {serverUrl: $u}' "$mcp_cfg" >"$tmp"
+      if jq --arg u "$server_url" '.mcpServers.monk = {serverUrl: $u}' "$mcp_cfg" >"$tmp" 2>/dev/null; then
+        generated=1
+      fi
     else
-      jq -n --arg u "$server_url" '{mcpServers: {monk: {serverUrl: $u}}}' >"$tmp"
+      if jq -n --arg u "$server_url" '{mcpServers: {monk: {serverUrl: $u}}}' >"$tmp" 2>/dev/null; then
+        generated=1
+      fi
     fi
-  elif command -v python3 >/dev/null 2>&1; then
+  fi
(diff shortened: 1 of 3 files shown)


ahmedabdelkarim504-commits:
@nooga Hi! This PR fixes #12 — a malformed  caused jq to exit non-zero, killing the launcher after the health check passed. 

The fix wraps jq calls with error handling and falls through to python3 (using  instead of string interpolation to avoid apostrophe issues). If both fail, it prints manual registration instructions and continues. All 3 POSIX copies updated. Would appreciate a review when you can!

nooga:
Thanks for this — good fix, but it's identical to your earlier PR #78, and issue #12 already has a canonical fix in progress via PR #13. Closing this as a duplicate.

Issue #25, opened 2026-07-17
[Bug bounty] unrelated health resource is accepted as the Monk companion


Dillxn:
### Stage

Connect

### Coding agent

Codex

### Summary

The POSIX launcher treats any successful response containing a JSON key named `resource` as proof that the configured local endpoint is `monk-agent`:

```sh
curl -fsS --max-time 2 "$health_url" 2>/dev/null | grep -q '"resource"'
```

It never verifies the value. If another local MCP-style service is already listening on the configured host/port and its protected-resource document names a different endpoint, the launcher exits successfully without starting the Monk companion. The host then connects to the wrong or unusable service while the startup hook reports no error.

The PowerShell launcher has the same functional check with `$Response.Content -match '"resource"'`. I have not claimed a native Windows reproduction.

### Deterministic reproduction

The isolated regression is available here: [start-monk-agent-health-identity.sh](https://gist.github.com/Dillxn/cf6fb7c02feb3b6cef11188fd7ccd9be#file-start-monk-agent-health-identity-sh).

It uses:

- a temporary `HOME` and launcher state directory;
- a fake custom `monk-agent` that creates a marker only when it is actually started;
- a local `curl` stub that initially returns `{"resource":"http://127.0.0.1:9999/not-monk"}`;
- the correct `http://127.0.0.1:7419/mcp` resource only after the fake companion starts.

Run:

```sh
sh tests/start-monk-agent-health-identity.sh
```

### Actual

Released `origin/main` (`a2a3482`) exits 0 without starting the fake companion. The regression then exits 1 with:

```text
launcher accepted an unrelated health resource as monk-agent
```

### Expected

The launcher should reuse a healthy process only when the protected-resource document's top-level `resource` value equals its configured MCP URL (`http://$host:$port/mcp`). A response naming another resource should be rejected so the real companion is started; once the correct resource appears, the health check should pass.

### User impact

A stale development server, another local MCP service, or a custom port collision can make every session appear to start successfully while Monk tools remain unavailable or target the wrong endpoint. Because the launcher suppresses the mismatch and returns success, the visible symptom is a confusing MCP/connect failure rather than a clear port/identity error.

### Setup

- macOS 26.5.2 host with `uname` stubbed to Linux for the POSIX background-process path
- arm64
- Codex CLI 0.144.5
- Monk plugin 0.1.42
- No real account, network endpoint, or user configuration touched by the regression

### Severity

[...]

PR #86, opened 2026-07-20
fix: improve Antigravity POSIX hook health check and jq handling

### LINKED ISSUE #40: [Bug bounty] Antigravity hook ignores healthy wget-only companion

ahmedabdelkarim504-commits:
Three issues fixed in .antigravity-plugin/hooks/ensure-monk-agent.sh:

1. #40: is_running() now falls back to wget when curl is unavailable, matching the contract used by the normal launcher.

2. #24: jq calls are guarded with command -v checks. When jq is absent, printf generates equivalent JSON output. This prevents exit 127 on systems without jq.

3. #25: Health check now validates the resource value equals the expected MCP URL instead of accepting any response containing the string resource. This prevents accepting an unrelated local service as the Monk companion.

Fixes #40, #24, #25

### DIFF
--- a/.antigravity-plugin/hooks/ensure-monk-agent.sh
+++ b/.antigravity-plugin/hooks/ensure-monk-agent.sh
@@ -14,8 +14,19 @@ host="${MONK_AGENT_HOST:-127.0.0.1}"
 health_url="http://$host:$port/.well-known/oauth-protected-resource"
 
 is_running() {
-  command -v curl >/dev/null 2>&1 || return 1
-  curl -fsS --max-time 2 "$health_url" 2>/dev/null | grep -q '"resource"'
+  expected_resource="http://$host:$port/mcp"
+  health_body=""
+  if command -v curl >/dev/null 2>&1; then
+    health_body="$(curl -fsS --max-time 2 "$health_url" 2>/dev/null || true)"
+  elif command -v wget >/dev/null 2>&1; then
+    health_body="$(wget -q -T 2 -O - "$health_url" 2>/dev/null || true)"
+  else
+    return 1
+  fi
+  case "$health_body" in
+    *"$expected_resource"*) return 0 ;;
+  esac
+  return 1
 }
 
 # Fast path — already up
@@ -29,11 +40,15 @@ agent_path="${MONK_AGENT_PATH:-${MONK_AGENT_INSTALL_DIR:-"$HOME/.monk/bin"}/monk
 
 if [ ! -x "$agent_path" ]; then
   plugin_dir="$(CDPATH= cd -- "$(dirname -- "$0")/.." && pwd)"
-  jq -n --arg script "$plugin_dir/scripts/start-monk-agent.sh" '{
-    injectSteps: [{
-      ephemeralMessage: ("monk-agent is not installed. Run `" + $script + "` once to install and start it, then continue.")
-    }]
-  }'
+  if command -v jq >/dev/null 2>&1; then
+    jq -n --arg script "$plugin_dir/scripts/start-monk-agent.sh" '{
+      injectSteps: [{
+        ephemeralMessage: ("monk-agent is not installed. Run `" + $script + "` once to install and start it, then continue.")
+      }]
+    }'
+  else
+    printf '{"injectSteps":[{"ephemeralMessage":"monk-agent is not installed. Run `%s` once to install and start it, then continue."}]}\n' "$plugin_dir/scripts/start-monk-agent.sh"
+  fi
   exit 0
 fi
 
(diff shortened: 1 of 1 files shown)


ahmedabdelkarim504-commits:
@nooga Hi! This PR fixes 3 issues in {
  "injectSteps": [
    {
      "ephemeralMessage": "monk-agent is not installed. Run `E:/AiLab/monk-plugin/.antigravity-plugin/scripts/start-monk-agent.sh` once to install and start it, then continue."
    }
  ]
}:

- **#40**:  now falls back to  when  is unavailable
- **#24**: jq calls guarded with  checks — prevents  on systems without jq (falls back to printf)
- **#25**: Health check now validates the resource value equals the expected MCP URL instead of accepting any response

All fixes in the Antigravity hook copy. Ready for review!

nooga:
Thanks for this, but on closer review it doesn't cover the full scope of issue #25 — it only patches the Antigravity PreInvocation hook (.antigravity-plugin/hooks/ensure-monk-agent.sh), leaving the main SessionStart launcher (start-monk-agent.sh, all 3 rendered copies) and both PowerShell equivalents equally exploitable, and doesn't include a regression test. PR #132 covers all 5 vulnerable call sites with test coverage, so we're going with that one. Closing this — appreciate the earlier work on #24 though, which did ship.

PR #91, opened 2026-07-20
fix: reuse installed agent when update check fails

### LINKED ISSUE #90: [Bug bounty] cold start fails offline despite a verified local agent
### Bounty eligibility

## Summary

With the default `MONK_AGENT_AUTO_UPDATE=1`, every plugin launch fetches the remote `.sha256` sidecar **before** it checks whether a verified `monk-agent` is already installed locally. If `get.monk.io` is unavailable (offline laptop, DNS/TLS failure, captive portal, or service outage), the launcher exits immediately and never starts the existing binary.

This makes the installed plugin unnecessarily dependent on the update service for every cold start. The same ordering exists in both `ensure-monk-agent.ps1` and `ensure-monk-agent.sh`, so the failure affects Windows, macOS, and Linux.

## Environment

- Monk plugin: `0.1.45`
- monk-agent: `0.1.45`
- Host: Codex on Windows 11 x64
- A valid, previously downloaded binary and matching installed checksum were already present in `~/.monk/bin`

## Reproduction

1. Install and start the plugin once while online so `monk-agent` and `monk-agent.sha256` exist.
2. Stop the agent (or reboot so the next session needs a cold start).
3. Make the update endpoint unavailable. The following isolated reproduction uses a loopback port with no listener rather than changing the machine's network configuration:

```powershell
$env:MONK_AGENT_HOME = '<isolated-test-home>'
$env:MONK_AGENT_INSTALL_DIR = "$HOME\.monk\bin"
$env:MONK_AGENT_DOWNLOAD_BASE = 'http://127.0.0.1:1'
$env:MONK_AGENT_PORT = '7420'
& '<plugin-root>\scripts\start-monk-agent.ps1'
```

4. Observe that the launcher exits before it attempts to start the already-installed agent:

```text
Invoke-WebRequest : Unable to connect to the remote server
At ...\scripts\ensure-monk-agent.ps1:105 char:1
+ Invoke-WebRequest -Uri $ChecksumUrl -OutFile $ChecksumTmp
```

The process exits with code `1`, and no agent listens on the requested port.

## Why it happens

The Windows bootstrap performs these operations in this order:

1. Download `$ChecksumUrl` to `$ChecksumTmp`.
2. Read the expected remote checksum.
3. Only then compare the installed checksum and return the existing target.

The POSIX bootstrap does the same with `curl`/`wget` before checking `target` plus `checksum_installed`.

Because `start-monk-agent.*` invokes the ensure script before its health/start path, an update-check failure aborts the complete cold-start lifecycle.

## Actual result

An installed, previously verified agent cannot start when the update endpoint is temporarily unavailable. Monk MCP is therefore unavailable for that session even though every required local artifact is present.

## Expected result

The launcher should start the installed verified agent when the update check cannot be completed, while surfacing the update failure as a non-fatal warning. A remote checksum mismatch should still trigger the normal verified update path when the endpoint is reachable.

## Suggested fix

Treat update discovery as best-effort when a target plus installed checksum already exist:

- attempt to fetch the remote checksum;
- if the fetch succeeds, keep the current comparison/update behavior;
- if it fails and a locally checksummed binary exists, warn and return that binary;
- remain fail-closed when no valid local installation exists.

This preserves first-install integrity while making subsequent cold starts resilient to transient update-service failures.

nexicturbo:
## Summary

- fall back to a non-empty, previously checksummed local `monk-agent` when update discovery is temporarily unavailable
- keep first installs fail-closed when no valid local installation exists
- apply the behavior consistently to the root, packaged, and Antigravity PowerShell/POSIX scripts
- exercise the offline fallback on Windows, macOS, and Linux in the existing install E2E workflow

## Why

The ensure scripts currently fetch the remote archive checksum before consulting the installed binary and checksum sidecar. A cold start therefore aborts during any transient DNS, TLS, captive-portal, network, or `get.monk.io` failure, even after a successful prior installation.

The fallback only applies when the target is non-empty and the existing sidecar contains a valid 64-character SHA-256 value. An unavailable update endpoint still causes a hard failure when no previously checksummed local installation exists.

## Validation

- reproduced the original Windows failure with an installed v0.1.45 agent and an unreachable update endpoint
- verified the updated PowerShell bootstrap returns the installed agent with the endpoint unavailable
- verified an empty install directory still exits non-zero under the same outage
- verified the normal online checksum/update path still returns the installed agent
- verified all three distributed PowerShell copies and all three POSIX copies are byte-identical
- parsed `.github/workflows/install-e2e.yml` and ran `git diff --check`

Fixes #90

### DIFF
--- a/.antigravity-plugin/scripts/ensure-monk-agent.ps1
+++ b/.antigravity-plugin/scripts/ensure-monk-agent.ps1
@@ -111,7 +111,31 @@ $ChecksumTmp = Join-Path $InstallDir ".monk-agent.tmp.sha256"
 $ExtractDir = Join-Path $InstallDir ".monk-agent.extract"
 New-Item -ItemType Directory -Force -Path $InstallDir | Out-Null
 
-Invoke-WebRequest -Uri $ChecksumUrl -OutFile $ChecksumTmp
+try {
+  Invoke-WebRequest -Uri $ChecksumUrl -OutFile $ChecksumTmp
+} catch {
+  $Installed = ""
+  if ((Test-Path $Target) -and (Test-Path $ChecksumInstalled)) {
+    try {
+      $Installed = ((Get-Content -Raw $ChecksumInstalled).Trim() -split "\s+")[0].ToLowerInvariant()
+    } catch {
+      $Installed = ""
+    }
+  }
+
+  $TargetSize = if (Test-Path $Target) { (Get-Item $Target).Length } else { 0 }
+  $ActualInstalled = if ($TargetSize -gt 0) { Get-FileSha256 $Target } else { "" }
+  if ($TargetSize -gt 0 -and
+      $Installed -match "^[0-9a-f]{64}$" -and
+      $ActualInstalled -eq $Installed) {
+    Remove-Item -Force $ChecksumTmp -ErrorAction SilentlyContinue
+    Write-Warning "Unable to check for monk-agent updates; using the previously checksummed installation at $Target."
+    Write-Output $Target
+    exit 0
+  }
+
+  throw
+}
 
 $Expected = ((Get-Content -Raw $ChecksumTmp).Trim() -split "\s+")[0].ToLowerInvariant()
 
--- a/.antigravity-plugin/scripts/ensure-monk-agent.sh
+++ b/.antigravity-plugin/scripts/ensure-monk-agent.sh
@@ -73,13 +73,41 @@ if [ "$auto_update" = "0" ] || [ "$auto_update" = "false" ]; then
   fi
 fi
 
-if command -v curl >/dev/null 2>&1; then
-  curl -fL "$checksum_url" -o "$checksum_tmp"
-elif command -v wget >/dev/null 2>&1; then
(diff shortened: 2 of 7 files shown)


nexicturbo:
Rebased onto v0.1.46 (a4c9972). The three POSIX copies and three PowerShell copies remain byte-identical by platform; shell syntax, PowerShell AST parsing, and git diff --check pass. Ready for maintainer workflow approval/review.

nexicturbo:
Rebased onto current `v0.1.49` (`64e8a0f`) and revalidated the updated branch. PowerShell AST parsing, POSIX shell syntax, workflow YAML parsing, distributed-script identity checks, and `git diff --check` all pass. The PR remains mergeable and is ready for maintainer review.

nooga:
Closing as already shipped — your fix was ported into the plugin via monk-agent's ENG-422 work (commit `ac9fe06`), which credits this PR directly. The offline checksum fallback is now in both the `.sh` and `.ps1` launchers with regression tests.

Correct diagnosis and correct fix; that's recorded in triage. Thanks.

PR #101, opened 2026-07-21
fix: preserve shell guard on helper failure

### LINKED ISSUE #100: [Bug bounty] Windows block-monk guard fails open when helper errors
## Bounty eligibility

- [x] I have signed up at monk.io with this GitHub account (`nexicturbo`)
- [x] I installed the plugin and reproduced this against the Windows hook path
- [x] I have starred the repository

## Stage

operate

## Coding agent

OpenAI Codex 0.125.0

## Plugin / agent version

0.1.45

## Repro steps

On Windows, `hooks/block-monk.ps1` delegates every Bash command to the installed `monk-agent hook block-monk` helper. If that helper exists but exits unsuccessfully, the script immediately exits with the helper's code instead of running its native fallback.

The following uses a stock Windows executable as a deterministic stand-in for an installed helper that fails during invocation. It does not touch a Monk cluster:

```powershell
$env:MONK_AGENT_PATH = "$env:SystemRoot\System32\net.exe"
$payload = '{"tool_input":{"command":"monk cluster list"}}'
$output = $payload |
    -File .\hooks\block-monk.ps1 2>&1

[pscustomobject]@{
  ExitCode = $LASTEXITCODE
  HasDenyDecision = (($output -join "`n") -match 'permissionDecision')
  Output = ($output -join "`n")
}
```

On v0.1.45 this returns:

```text
ExitCode       : 1
HasDenyDecision: False
Output         : The syntax of this command is: ...
```

There is no deny JSON, so the PreToolUse host receives no block signal for the direct `monk` command. The same fail-open state can occur during an interrupted update, with an incompatible helper, or when the helper fails to initialize.

A second variant is an existing executable that returns success but emits malformed output. Before the fix, that output is relayed and trusted without validation, again bypassing the native command check.

## Expected behavior

The shell guard should keep blocking a direct `monk` CLI invocation whenever the helper cannot return a valid PreToolUse decision. Ordinary commands must remain allowed and produce no output.

## Actual behavior

Any existing helper path bypasses the PowerShell fallback. A nonzero helper result is propagated, while a successful malformed response is relayed as if it were a valid decision. Neither path produces the deny JSON that blocks the direct CLI invocation.

This defeats the guard precisely when the installed companion is partially updated, incompatible, or otherwise unhealthy. Running the CLI directly can desynchronize the cluster state that Monk manages, which is the failure the hook exists to prevent.

## Proposed fix

Capture and validate the helper response. Trust it only when it exits successfully and returns a structured `PreToolUse` permission decision; otherwise discard its output and run the native PowerShell parser. A Windows regression should cover both a nonzero helper and a successful helper with malformed output, for blocked and allowed commands.

## Severity (my guess)

major

## Environment

- OS: Windows 11 Home 64-bit, build 26200
- PowerShell: 5.1.26100.8894
- monkd: not involved; the failure occurs in the host hook before runtime access
- Target cloud: none
- Integration: Claude Code-compatible Windows PreToolUse hook

nexicturbo:
## Summary

- fall back to the native PowerShell command parser when `monk-agent hook block-monk` fails
- validate successful helper output before treating it as an authoritative PreToolUse decision
- suppress broken helper output so ordinary non-Monk commands remain unaffected
- add Windows CI coverage for nonzero and malformed-success helper behavior

## Root cause

`hooks/block-monk.ps1` returned immediately whenever the configured helper path existed. A helper failure propagated its exit code without producing deny JSON, and a successful but malformed response was relayed without validation. In both cases, direct `monk` commands bypassed the native fallback that is supposed to preserve the shell guard when the companion is unhealthy.

## Impact

The guard now remains active through interrupted updates, incompatible helpers, and initialization failures. Direct `monk` commands are denied while ordinary commands still exit cleanly with no hook output.

## Validation

- reproduced the original nonzero-helper failure with the stock Windows `net.exe`
- exercised malformed-success output with `cmd.exe`
- verified both cases deny `monk cluster list` and allow `echo monk`
- verified the installed `monk-agent` v0.1.45 still denies/allows the same cases
- `git diff --check`

Closes #100

### DIFF
--- a/hooks/block-monk.ps1
+++ b/hooks/block-monk.ps1
@@ -8,21 +8,77 @@
 $agentDir = if ($env:MONK_AGENT_INSTALL_DIR) { $env:MONK_AGENT_INSTALL_DIR } else { Join-Path $HOME ".monk\bin" }
 $agent = if ($env:MONK_AGENT_PATH) { $env:MONK_AGENT_PATH } else { Join-Path $agentDir "monk-agent.exe" }
 
-# Let the binary read the hook payload straight from stdin. Reading it into a
-# PowerShell string first and re-piping it corrupts non-ASCII bytes (e.g. the
-# leading UTF-8 BOM that some hosts prepend) whenever the console is on an OEM
-# code page - which is exactly what the cmd.exe launcher sets - so the binary
-# would receive unparseable JSON and silently allow the command. The binary
-# reads raw UTF-8 (BOM-tolerant) from the inherited stdin and is unaffected.
+# Buffer stdin as bytes so the helper receives the original UTF-8 payload
+# without a PowerShell console-code-page round trip. Keeping the bytes also
+# lets the native parser make the decision if the helper fails or emits an
+# invalid response.
+$inputStream = [Console]::OpenStandardInput()
+$inputBuffer = New-Object byte[] 4096
+$payloadStream = New-Object System.IO.MemoryStream
+while (($bytesRead = $inputStream.Read($inputBuffer, 0, $inputBuffer.Length)) -gt 0) {
+  $payloadStream.Write($inputBuffer, 0, $bytesRead)
+}
+$hookBytes = $payloadStream.ToArray()
+$payloadStream.Dispose()
+
 if (Test-Path $agent) {
-  & $agent hook block-monk --format claude
-  exit $LASTEXITCODE
+  # Treat the helper as authoritative only when it succeeds and emits a
+  # decision. An interrupted update, incompatible binary, or startup failure
+  # must not turn the guard off: discard the helper error and use the native
+  # parser below. A successful empty response also falls through safely; the
+  # fallback permits ordinary commands and still blocks direct `monk` calls.
+  $agentProcess = $null
+  $agentText = $null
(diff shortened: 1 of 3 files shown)


nexicturbo:
Rebased onto v0.1.46 (a4c9972). The consolidated tests/block-monk-windows.ps1 regression passes locally for a missing helper, a failing helper (net.exe), and malformed-success output (cmd.exe), while preserving the shipped command-boundary cases. Ready for maintainer workflow approval/review.

nexicturbo:
Rebased onto the current `v0.1.49` branch and reconciled the newer Windows hook implementation without weakening the guard. The resolved path buffers raw stdin bytes, preserves UTF-8/BOM payloads when invoking the helper through `ProcessStartInfo`, and still falls back to the native parser if the helper fails or returns invalid output.

Fresh validation:
- `tests/block-monk-windows.ps1` passes
- the installed `monk-agent` hook correctly denies a BOM/non-ASCII test payload
- `git diff --check` passes

Current head: `5d0411ce042b2bb73059687451ec458409e53507`. The PR is mergeable and ready for maintainer review.

Lachann:

PR #104, opened 2026-07-21
fix: verify agent identity before stop

### LINKED ISSUE #98: [Bug bounty] stale POSIX agent PID can terminate an unrelated process
### Bounty eligibility

### Stage

install

### Coding agent

OpenAI Codex

### Coding agent version

Codex desktop 26.715.4045; bundled Codex CLI 0.145.0-alpha.18; Monk plugin 0.1.45

### Summary

The POSIX startup and uninstall scripts trust the numeric PID stored in `~/.monk/agent/launcher/run/monk-agent.pid` and send it `SIGTERM` without verifying that the PID still belongs to the managed `monk-agent` executable.

PID files outlive processes after crashes, forced exits, host shutdowns, and other unclean stops. Linux eventually reuses the PID. When that happens, the next coding-agent session runs `start-monk-agent.sh` automatically and can terminate the unrelated process now holding that PID. `uninstall-monk-agent.sh --yes` has the same behavior.

The data-loss impact depends on which process receives the reused PID: an editor, database, build, or other long-running user process can be terminated without warning merely by starting a coding-agent session or uninstalling the plugin.

### Repro steps

On Linux with the plugin and agent installed:

```sh
set -eu

export MONK_AGENT_HOME="$(mktemp -d)"
export MONK_AGENT_INSTALL_DIR="$HOME/.monk/bin"
export MONK_AGENT_PORT=17419
export MONK_AGENT_SKIP_ENSURE=1
export MONK_AGENT_SKIP_SIGNIN_NUDGE=1

run_dir="$MONK_AGENT_HOME/agent/launcher/run"
mkdir -p "$run_dir"

# Stand-in for an unrelated process that inherited a PID formerly used by the
# agent after the stale PID file was left behind.
sleep 300 &
victim_pid="$!"
printf '%s\n' "$victim_pid" >"$run_dir/monk-agent.pid"

./scripts/start-monk-agent.sh

# On current main this fails because the launcher sent SIGTERM to sleep.
kill -0 "$victim_pid"
```

The responsible startup code is:

```sh
old_pid="$(cat "$pid_file" 2>/dev/null || true)"
if [ -n "$old_pid" ]; then
  kill "$old_pid" >/dev/null 2>&1 || true
fi
```

The uninstaller independently repeats the same trust pattern after checking only that the file contains digits.

### Expected behavior

The launcher and uninstaller must verify that a PID still belongs to the managed `monk-agent` executable before signaling it. If identity cannot be proven, they should leave the process alone, discard or replace the stale PID record, and let the agent's existing port-conflict handling determine whether another healthy agent owns the port.

### Actual behavior

Any live process whose PID is written in the persistent PID file receives `SIGTERM` during POSIX startup and uninstall, regardless of its executable identity.

### Suggested fix

On Linux, resolve `/proc/<pid>/exe` and compare it with the canonical installed agent path before sending a signal. Treat missing `/proc`, malformed PIDs, unreadable process identity, and non-matching executable paths as a stale record and do not signal. Add regression coverage with a long-running unrelated process whose PID is placed in the stale file for both startup and uninstall.

### Severity (your guess)

major (unintended termination of an arbitrary unrelated user process, with possible data loss)

### OS

Linux; POSIX launcher and uninstaller

### monkd version

Not applicable; this occurs before/while managing the local companion process

### Target cloud

None; local plugin lifecycle

### Integration (if relevant)

Monk plugin SessionStart launcher and uninstall hook

### LINKED ISSUE #103: [Bug bounty] Windows PID guard trusts executable basename instead of managed path
### Bounty eligibility

### Stage

install

### Coding agent

OpenAI Codex

### Coding agent version

0.125.0

### Repro steps

1. On Windows, install and run `monk-agent.exe` from one directory. I used the real v0.1.45 binary installed at `C:\Users\Turbo\.monk\bin\monk-agent.exe` with an isolated `MONK_AGENT_HOME` and port.
2. Set `MONK_AGENT_INSTALL_DIR` to a different directory that would be managed by the plugin.
3. Put the running process's PID in that managed home's `agent\launcher\run\monk-agent.pid`.
4. Run `scripts\uninstall-monk-agent.ps1 -Yes -KeepData` from the current main branch. The same stop predicate is also present in the Windows start and ensure scripts.

I reproduced the predicate with a live process whose executable path was `C:\Users\Turbo\.monk\bin\monk-agent.exe` while the managed target was a separate `...\managed-bin\monk-agent.exe`. The health endpoint responded, the full paths were unequal, but the current basename check classified the process as safe to stop.

### Expected behavior

A PID file should authorize stopping a process only when that process is the agent executable at the plugin's managed install path. A stale or reused PID that points to an executable elsewhere should be discarded without stopping that process.

### Actual behavior

The Windows scripts trust the executable basename `monk-agent.exe` rather than the managed executable's full path. The uninstaller also treats an unreadable process path as managed. Consequently, a stale PID can make install, start, or uninstall forcibly stop a differently located agent process.

I tested a fix that compares normalized full paths case-insensitively and fails closed when the path is empty or unreadable. In an end-to-end test using the actual fixed uninstaller, the differently located agent remained alive and the stale PID file was removed.

### Severity (your guess)

major

### OS

Windows 11 build 26200; Windows PowerShell 5.1.26100.8894

### monkd version

monk-agent v0.1.45; plugin 0.1.45

### Target cloud

N/A (local companion lifecycle)

### Integration (if relevant)

Windows PowerShell lifecycle scripts

nexicturbo:
## Summary

- stop a PID-file process on Windows only when its normalized executable path matches the managed agent path
- on Linux, compare `/proc/<pid>/exe` with the canonical managed-agent path before startup or uninstall sends a signal
- preserve differently located, malformed, unreadable, and non-Linux stale-PID processes when identity cannot be proven
- keep root and packaged lifecycle scripts in sync
- add Windows and Linux end-to-end regressions for stale and managed PIDs

## Why

The lifecycle scripts trusted too little process identity before sending a stop signal. Windows checked only the executable basename, while the POSIX startup path trusted a numeric PID and the v0.1.46 uninstaller accepted any process named `monk-agent`. A stale or reused PID could therefore terminate a process outside the plugin's managed installation, including a differently located `monk-agent` on Linux or Windows.

The Windows helper now performs a case-insensitive normalized full-path comparison. The Linux helper resolves `/proc/<pid>/exe` and the expected executable path, and fails closed when either identity cannot be verified. Normal managed-agent shutdown remains covered in the regression tests.

Fixes #98.
Fixes #103.

## Validation

- parsed all six changed PowerShell scripts with the PowerShell AST parser
- exercised the Windows path helper in every root and packaged copy: case-insensitive equivalent paths match; differently located same-basename paths and empty paths do not
- ran the fixed Windows uninstaller end to end with a live real `monk-agent.exe` from a different directory; the process survived and the stale PID file was removed
- checked both changed POSIX scripts with Bash 5.3 syntax validation
- exercised the Linux identity helper against matching, mismatching, malformed, and deleted-executable path cases
- added Linux CI coverage for startup and uninstall with an unrelated stale PID, plus uninstall of the actual managed agent
- confirmed each packaged lifecycle script is byte-identical to its root counterpart
- `git diff --check`

### DIFF
--- a/.antigravity-plugin/hooks/block-monk.ps1
+++ b/.antigravity-plugin/hooks/block-monk.ps1
@@ -35,7 +35,15 @@ $hookInput = $reader.ReadToEnd()
 try { $command = ($hookInput | ConvertFrom-Json).toolCall.args.CommandLine } catch { exit 0 }
 if (-not $command) { exit 0 }
 
-if ($command -match '(^|[\r\n;&|`({])\s*(sudo\s+)?monk(\s|$)') {
+# Shell quoting/escaping ("monk", m\onk) and a short wrapper-command list
+# don't change what actually runs, so strip backslashes/quotes before
+# matching and recognize `monkd` + a leading forward-slash path. Blunt,
+# non-quote-aware strip — known gap versus the primary `monk-agent hook
+# block-monk` path: a backslash-separated Windows path loses its separator
+# to the strip here and isn't detected (see plugin/static/claude/hooks/
+# block-monk.ps1 for the same tradeoff, spelled out in more detail).
+$normalized = $command.Replace('\', '').Replace('"', '').Replace("'", '')
+if ($normalized -match '(^|[\r\n;&|`({])\s*(sudo|command|env|exec|powershell(\.exe)?\s+-(Command|c)|cmd(\.exe)?\s+/c)?\s*([^\s;&|`(){}]*[\\/])?monkd?(\.(exe|cmd|bat|ps1))?(\s|$)') {
   @{
     decision = "deny"
     reason   = "Blocked: do not shell out to the ``monk`` CLI - it desyncs the cluster state Monk manages. Use the monk-agent MCP tools instead."
--- a/.antigravity-plugin/hooks/block-monk.sh
+++ b/.antigravity-plugin/hooks/block-monk.sh
@@ -30,9 +30,29 @@ if [ -x "$agent" ]; then
   fi
 fi
 
-# Fallback: binary unavailable. Grep the raw hook payload for a `monk` command
-# in command position. False positives only ever BLOCK, never allow.
-if printf '%s' "$input" | grep -Eq '(^|["[:space:];&|`(])(sudo[[:space:]]+)?monk([[:space:]"]|$)'; then
+# Fallback: binary unavailable. No jq here, so pull just the `CommandLine`
+# string out of the raw JSON (best-effort key match, not a full parser) rather
+# than grepping the whole payload: matching against the raw text means a `"`
+# right before "monk" is the JSON *string* delimiter, not shell quoting —
+# stripping quotes from the whole payload destroys that boundary. Once
+# isolated, JSON's own `\"`/`\\` escaping and shell-level quoting/escaping
+# ("monk", m\onk) both collapse under the same blunt backslash/quote strip
+# (they compose additively here). Then match `monk`/`monkd` in command
+# position, past a short wrapper-command list and an optional leading path.
+# This is a blunt, non-quote-aware strip — see block-monk.ps1 for the
+# tradeoff vs the quote-state-aware primary `monk-agent hook block-monk` path.
+# False positives only ever BLOCK, never allow.
+raw_command="$(printf '%s' "$input" |
+  grep -Eo '"CommandLine"[[:space:]]*:[[:space:]]*"([^"\\]|\\.)*"' |
+  sed -E 's/^"CommandLine"[[:space:]]*:[[:space:]]*"//; s/"$//')"
+# A real newline in the original command is JSON-encoded as the two-char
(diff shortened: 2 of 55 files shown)


nexicturbo:
Rebased onto v0.1.46 (a4c9972) and narrowed the PR description to #98/#103 because v0.1.46 already shipped the #53 basename safeguard. Shell syntax, all six PowerShell AST parses, git diff --check, and every root/package blob-pair check pass. Ready for maintainer workflow approval/review.

nexicturbo:
Rebased both commits onto current `v0.1.49` (`c8bd520`) and revalidated the updated branch. PowerShell AST parsing, POSIX shell syntax, workflow YAML parsing, distributed-script identity checks, and `git diff --check` all pass. The PR remains mergeable and is ready for maintainer review.

Lachann:

PR #106, opened 2026-07-21
fix: harden managed agent bootstrap

### LINKED ISSUE #105: [Bug bounty] Windows disabled-update bootstrap prefers PATH agent over managed install
### Bounty eligibility

### Stage

install

### Coding agent

OpenAI Codex

### Coding agent version

0.125.0

### Repro steps

1. On Windows, install the real `monk-agent.exe` into an isolated managed directory.
2. Put a second `monk-agent.exe` in another directory and prepend that directory to `PATH`.
3. Set `MONK_AGENT_INSTALL_DIR` to the managed directory and set `MONK_AGENT_AUTO_UPDATE=0`.
4. Run `scripts\ensure-monk-agent.ps1` from current `main`.

I reproduced this with two copies of the installed v0.1.45 executable. The managed target was:

```text
...\managed-bin\monk-agent.exe
```

The separate executable on `PATH` was:

```text
...\external-bin\monk-agent.exe
```

The bootstrap returned the external path, with `ReturnedExternal=true`, even though the managed target already existed.

The Windows script checks `Get-Command monk-agent.exe` before `$Target` in its disabled-update branch. The POSIX bootstrap has the intended opposite order and prefers its managed target first.

### Expected behavior

When the managed agent exists and updates are disabled, the bootstrap should return that managed executable. A `PATH` lookup should be only a fallback when the managed target is absent.

### Actual behavior

Any earlier `monk-agent.exe` on `PATH` wins over the plugin's existing managed installation. The launcher then uses that other executable, which can be an unintended version or installation, while the bootstrap reports success.

I tested a three-file fix that reverses the Windows lookup order in the root and packaged scripts and adds a Windows end-to-end regression. The managed-present case now returns the managed target, while the managed-missing case still returns the `PATH` fallback.

### Severity (your guess)

major

### OS

Windows 11 build 26200; Windows PowerShell 5.1.26100.8894

### monkd version

monk-agent v0.1.45; plugin 0.1.45

### Target cloud

N/A (local companion bootstrap)

### Integration (if relevant)

Windows PowerShell installer

### LINKED ISSUE #109: [Bug bounty] Windows launcher mutex is port-scoped, allowing shared installer race
### Stage

install

### Coding agent

OpenAI Codex

### Coding agent version

Codex desktop; Monk plugin v0.1.45

### Repro steps

Check out `monk-io/monk-plugin` at `1e14687724cf5dbcf5a2ac80339d3927d3922d60` (v0.1.45) on Windows 11.

Use a fresh isolated `MONK_AGENT_INSTALL_DIR`, then start two copies of `scripts/start-monk-agent.ps1` at the same time with:

- the same `MONK_AGENT_INSTALL_DIR`
- separate `MONK_AGENT_HOME` directories
- different valid `MONK_AGENT_PORT` values (I used `17431` and `17432`)

For each process I used `powershell.exe -NoProfile -ExecutionPolicy Bypass -File scripts\start-monk-agent.ps1`, redirected stdout/stderr to separate files, and waited for both processes.

The two launchers create different named mutexes:

```text
Local\monk-agent-launcher-17431
Local\monk-agent-launcher-17432
```

but both invoke `ensure-monk-agent.ps1` against these same process-shared paths:

```text
$MONK_AGENT_INSTALL_DIR\.monk-agent.tmp.sha256
$MONK_AGENT_INSTALL_DIR\.monk-agent.tmp.zip
$MONK_AGENT_INSTALL_DIR\.monk-agent.extract
```

A clean run against the real stable Windows artifact returned one success and one failure:

```text
exit_codes=0,1
Invoke-WebRequest: The process cannot access the file
'...\install\.monk-agent.tmp.zip' because it is being used by another process.
At scripts\ensure-monk-agent.ps1:135
```

A direct two-process bootstrap also failed immediately on `.monk-agent.tmp.sha256`.

I additionally pinned a network-free forced-overlap check to the exact public commit above. Two PowerShell jobs started together and replaced `Invoke-WebRequest` with a local fixture writer that opens `.monk-agent.tmp.sha256` with `FileShare.None` for 750 ms. The unchanged source produced:

```text
states=Failed,Completed
errors=1
successful_paths=1
IOException: ...\.monk-agent.tmp.sha256 is being used by another process
```

This is Windows-specific and separate from #18: that report exercises `ensure-monk-agent.sh` on Unix, while this failure is in the independent PowerShell installer and bypasses the Windows launcher's existing mutex specifically because that mutex is scoped to the agent port.

### Expected behavior

Concurrent Windows launchers that share the managed installation should serialize the installation transaction regardless of their agent ports. Both should complete with the same verified executable.

### Actual behavior

Different ports create different launcher mutexes, so the PowerShell installers concurrently overwrite, extract, move, and delete the same scratch paths. One legitimate launcher exits non-zero and plugin startup aborts.

A mutex inside `ensure-monk-agent.ps1`, covering the complete checksum/download/extract/move transaction independently of the port, prevents both launcher and direct-bootstrap races.

This is a reproducible non-security defect.

### Severity (your guess)

major

### OS

Windows 11 24H2 x64

### monkd version

Not installed or invoked; the failure occurs in the companion agent bootstrap.

### Target cloud

None; isolated local install directory.

### Integration (if relevant)

Monk plugin bootstrap from simultaneous OpenAI Codex sessions using separate agent ports.

nexicturbo:
## Summary

- prefer the plugin-managed `monk-agent` when automatic updates are disabled on Windows, Linux, and macOS
- retain the existing `PATH` lookup as a fallback when the managed target is absent
- serialize the complete Windows checksum/download/extract/move transaction independently of the agent port
- add end-to-end regressions for managed-path precedence and concurrent PowerShell installers

## Why

The disabled-update branches checked `PATH` before their managed targets. An unrelated or older executable earlier on `PATH` therefore won even when the plugin's managed installation existed.

Separately, the Windows launcher's mutex is scoped to `MONK_AGENT_PORT`, but every launcher shares fixed installer scratch paths. Launchers using different valid ports therefore bypassed each other's mutex and raced on the checksum, archive, and extraction directory. One legitimate startup failed with a file-in-use error.

Fixes #105.
Fixes #109.

## Validation

- reproduced #109 on exact public commit `1e14687724cf5dbcf5a2ac80339d3927d3922d60` with a forced two-process overlap: one failed and one completed
- reproduced the real launcher path with ports `17431` and `17432`: one exited `1` on the shared archive and one exited `0`
- verified the fixed deterministic regression returns the managed executable from both installers with zero errors
- verified a fixed cold install against the real 65 MB stable Windows artifact: both installers exited `0`, the executable and checksum were present, and no scratch paths remained
- reproduced the Windows managed-vs-`PATH` precedence behavior with the real v0.1.45 executable
- verified the fixed managed-present case returns the managed target and the managed-missing case retains the `PATH` fallback
- added the equivalent managed-path regression to the Ubuntu/macOS install matrix
- parsed both changed PowerShell scripts with the PowerShell AST parser
- confirmed the root and packaged bootstrap scripts remain identical
- `git diff --check`

### DIFF
--- a/.github/workflows/install-e2e.yml
+++ b/.github/workflows/install-e2e.yml
@@ -38,6 +38,21 @@ jobs:
           test "$installed" = "$MONK_AGENT_INSTALL_DIR/monk-agent"
           test -x "$installed"
 
+      - name: Prefer managed agent when updates are disabled
+        shell: bash
+        env:
+          MONK_AGENT_INSTALL_DIR: ${{ runner.temp }}/monk-bin
+          MONK_AGENT_AUTO_UPDATE: "0"
+        run: |
+          set -euo pipefail
+          other_dir="$RUNNER_TEMP/other-monk-bin"
+          mkdir -p "$other_dir"
+          cp "$MONK_AGENT_INSTALL_DIR/monk-agent" "$other_dir/monk-agent"
+          export PATH="$other_dir:$PATH"
+
+          installed="$(./scripts/ensure-monk-agent.sh)"
+          test "$installed" = "$MONK_AGENT_INSTALL_DIR/monk-agent"
+
       - name: Run monk-agent status
         shell: bash
         env:
--- a/.github/workflows/install-e2e.yml
+++ b/.github/workflows/install-e2e.yml
@@ -118,6 +118,60 @@ jobs:
             throw "disabled-update bootstrap selected PATH agent: $installed"
           }
 
+      - name: Serialize concurrent Windows installers
+        shell: pwsh
+        env:
+          MONK_AGENT_INSTALL_DIR: ${{ runner.temp }}\monk-bin
+        run: |
+          $script = (Resolve-Path ".\scripts\ensure-monk-agent.ps1").Path
+          $checksum = "$env:MONK_AGENT_INSTALL_DIR\monk-agent.sha256"
+          $gate = "$env:RUNNER_TEMP\monk-installer-gate-$([Guid]::NewGuid())"
+
+          $jobs = 1..2 | ForEach-Object {
+            Start-Job -ArgumentList $script, $checksum, $gate, $env:MONK_AGENT_INSTALL_DIR -ScriptBlock {
+              param($script, $checksum, $gate, $installDir)
+              $ErrorActionPreference = "Stop"
+              $env:MONK_AGENT_INSTALL_DIR = $installDir
+
+              New-Item -ItemType File -Path "$gate-$PID" | Out-Null
+              while ((Get-ChildItem "$gate-*" -ErrorAction SilentlyContinue).Count -lt 2) {
+                Start-Sleep -Milliseconds 20
+              }
+
+              function Invoke-WebRequest {
+                param([string]$Uri, [string]$OutFile)
+                $bytes = [IO.File]::ReadAllBytes($checksum)
+                $stream = [IO.File]::Open(
+                  $OutFile,
+                  [IO.FileMode]::Create,
+                  [IO.FileAccess]::Write,
+                  [IO.FileShare]::None
+                )
+                try {
+                  $stream.Write($bytes, 0, $bytes.Length)
+                  $stream.Flush()
+                  Start-Sleep -Milliseconds 750
+                } finally {
+                  $stream.Dispose()
+                }
+              }
+
+              & $script
+            }
+          }
+
+          $jobs | Wait-Job | Out-Null
+          $errors = @()
+          $results = $jobs | Receive-Job -ErrorVariable +errors
+          $states = $jobs.State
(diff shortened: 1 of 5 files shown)


nexicturbo:
Rebased onto v0.1.46 (a4c9972) without conflict. Root/package bootstrap blobs remain identical; POSIX syntax, PowerShell AST parsing, and git diff --check pass. Ready for maintainer workflow approval/review.

nexicturbo:
Rebased all three commits onto current `v0.1.49` (`95bc84b`) and revalidated the updated branch. PowerShell AST parsing, POSIX shell syntax, workflow YAML parsing, distributed-script identity checks, and `git diff --check` all pass. The PR remains mergeable and is ready for maintainer review.

Lachann:

PR #111, opened 2026-07-21
fix: propagate plugin version on Windows

### LINKED ISSUE #110: [Bug bounty] Windows launcher omits rendered plugin version from agent telemetry
### Stage

Connect

### Coding agent

OpenAI Codex

### Coding agent version

codex-cli 0.125.0; Monk plugin v0.1.45 (`1e14687724cf5dbcf5a2ac80339d3927d3922d60`)

### Repro steps

On native Windows, `scripts/start-monk-agent.ps1` launches the companion without loading the build-rendered `scripts/plugin-version.sh`. The Unix launcher sources that file explicitly so `MONK_PLUGIN_VERSION` reaches the child process, but the PowerShell launcher sets only the auth/autospin variables before `Start-Process`.

This deterministic reproduction runs the shipped Windows launcher against a temporary capture process. It neither installs Monk nor contacts Monk services:

```powershell
$root = Join-Path $env:TEMP "monk-plugin-version-repro"
New-Item -ItemType Directory -Force -Path $root | Out-Null
$capture = Join-Path $root "capture.txt"
$fakeAgent = Join-Path $root "monk-agent.cmd"

@'
@echo off
if defined MONK_PLUGIN_VERSION (
  echo MONK_PLUGIN_VERSION=%MONK_PLUGIN_VERSION%>"%MONK_CAPTURE_PATH%"
) else (
  echo MONK_PLUGIN_VERSION=^<missing^>>"%MONK_CAPTURE_PATH%"
)
echo ARGS=%*>>"%MONK_CAPTURE_PATH%"
exit /b 0
'@ | Set-Content -Encoding ascii $fakeAgent

$env:MONK_AGENT_PATH = $fakeAgent
$env:MONK_AGENT_HOME = Join-Path $root "agent-home"
$env:MONK_AGENT_INSTALL_DIR = Join-Path $root "install"
$env:MONK_AGENT_PORT = "17499"
$env:MONK_AGENT_SKIP_ENSURE = "1"
$env:MONK_AGENT_SKIP_SIGNIN_NUDGE = "1"
$env:MONK_CAPTURE_PATH = $capture
[Environment]::SetEnvironmentVariable("MONK_PLUGIN_VERSION", $null, "Process")

  -File .\scripts\start-monk-agent.ps1 2>$null

Get-Content -Raw $capture
```

Observed on the exact public commit above:

```text
MONK_PLUGIN_VERSION=<missing>
ARGS=serve --host 127.0.0.1 --port 17499
```

The fake process exits immediately, so the launcher itself returns its expected not-ready error. The relevant observation is the exact environment inherited by the process at the real `Start-Process` boundary.

For comparison, `scripts/plugin-version.sh` contains:

```sh
MONK_PLUGIN_VERSION="0.1.45"
```

and `scripts/start-monk-agent.sh` sources and exports it before launching the same companion.

### Expected behavior

Native Windows launches should pass the build-rendered plugin version to `monk-agent`, matching the Unix launch path. Telemetry and diagnostics should identify the plugin release that started the companion.

### Actual behavior

`MONK_PLUGIN_VERSION` is absent from every native Windows companion launched by the plugin unless the caller happens to define it independently. The agent therefore falls back to its own binary version, making native Windows sessions indistinguishable across plugin releases that ship the same agent build.

### Severity (your guess)

Moderate. The companion still starts, but release attribution and version-specific field diagnostics are wrong for the plugin's native Windows path. This makes rollout regressions harder to isolate and undercounts or misattributes Windows usage by plugin version.

### OS

Windows 11, native PowerShell launcher

### monkd version

Not involved; the defect is at the companion process-launch boundary.

### Target cloud

None

### Integration (if relevant)

OpenAI Codex, Claude Code, and Antigravity native Windows packages generated from the shared PowerShell launcher.

nexicturbo:
## Summary

- load the build-rendered plugin version in the native Windows launcher
- pass `MONK_PLUGIN_VERSION` to the companion through its inherited process environment
- keep older rendered packages without `plugin-version.sh` backward compatible
- assert the Windows launcher exports the exact rendered version in install E2E CI

## Why

The POSIX launcher sources `plugin-version.sh` before starting `monk-agent`, but the PowerShell launcher did not. Native Windows companions therefore received no `MONK_PLUGIN_VERSION` and fell back to their agent-binary version, so telemetry could not attribute Windows sessions to the plugin release that launched them.

The fix parses only the generated `MONK_PLUGIN_VERSION="..."` assignment beside the launcher and exports it before the existing `Start-Process` call. All three rendered PowerShell copies remain identical.

Fixes #110.

## Impact

Native Windows usage and diagnostics now carry the same plugin-release identity as Unix launches. This makes rollout regressions and field reports attributable to the actual plugin version.

## Validation

- reproduced the unpatched launcher through its real `Start-Process` boundary: an isolated capture child received `MONK_PLUGIN_VERSION=<missing>`
- reran the same reproduction after the patch: the child received `MONK_PLUGIN_VERSION=0.1.45` and the unchanged `serve --host ... --port ...` arguments
- launched the real v0.1.45 Windows companion on an isolated port and home, verified HTTP 200 health metadata containing `resource`, then stopped the exact launched process
- parsed all three changed PowerShell launchers with the PowerShell AST parser (zero errors)
- verified the root, packaged Monk, and Antigravity PowerShell launchers are byte-identical
- `git diff --cached --check`

### DIFF
--- a/.antigravity-plugin/scripts/start-monk-agent.ps1
+++ b/.antigravity-plugin/scripts/start-monk-agent.ps1
@@ -8,6 +8,20 @@ $AuthAudience = if ($env:MONK_AUTH_AUDIENCE) { $env:MONK_AUTH_AUDIENCE } else { 
 $AutospinUrl = if ($env:MONK_AUTOSPIN_URL) { $env:MONK_AUTOSPIN_URL } else { "wss://api.app.monk.io/autospin/" }
 
 $ScriptDir = if ($PSScriptRoot) { $PSScriptRoot } else { Split-Path -Parent $MyInvocation.MyCommand.Path }
+# Rendered at plugin build time; carries MONK_PLUGIN_VERSION so the Windows
+# agent reports the plugin release rather than falling back to its binary
+# version. Guarded so older rendered plugins without the file still launch.
+$PluginVersionFile = Join-Path $ScriptDir "plugin-version.sh"
+if (Test-Path -LiteralPath $PluginVersionFile) {
+  $PluginVersionMatch = [regex]::Match(
+    (Get-Content -Raw -LiteralPath $PluginVersionFile),
+    '(?m)^MONK_PLUGIN_VERSION="([^"\r\n]+)"\s*$'
+  )
+  if ($PluginVersionMatch.Success) {
+    $env:MONK_PLUGIN_VERSION = $PluginVersionMatch.Groups[1].Value
+  }
+}
+
 $PluginRoot = if ($env:CLAUDE_PLUGIN_ROOT) { $env:CLAUDE_PLUGIN_ROOT } else { Split-Path -Parent $ScriptDir }
 $InstallDir = if ($env:MONK_AGENT_INSTALL_DIR) { $env:MONK_AGENT_INSTALL_DIR } else { Join-Path $HOME ".monk\bin" }
 $MonkHome = if ($env:MONK_AGENT_HOME) { $env:MONK_AGENT_HOME } else { Join-Path $HOME ".monk" }
--- a/.github/workflows/install-e2e.yml
+++ b/.github/workflows/install-e2e.yml
@@ -116,7 +116,16 @@ jobs:
           MONK_AGENT_SKIP_ENSURE: "1"
           MONK_TELEMETRY: "0"
         run: |
+          Remove-Item Env:MONK_PLUGIN_VERSION -ErrorAction SilentlyContinue
           .\scripts\start-monk-agent.ps1
+          $versionSource = Get-Content -Raw .\scripts\plugin-version.sh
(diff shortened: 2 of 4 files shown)


nexicturbo:
Implementation and local validation are complete. GitHub has disabled the fork author's **Ready for review** control while the workflow awaits maintainer approval. Please approve the workflow and mark this PR ready for review when convenient.

nooga:
Thanks for tracking this down — the underlying bug (missing MONK_PLUGIN_VERSION on native Windows launches) was already fixed internally shortly before you filed it, using a different approach. Since there's nothing left to merge here, we're closing this without a bounty score. We appreciate the investigation regardless.

PR #127, opened 2026-07-21
fix: bound launcher readiness by wall clock

### LINKED ISSUE #52: [Bug bounty] launcher readiness can exceed the host hook timeout

jamilahmadzai:
## Summary

- bound launcher readiness waits to 150 seconds so the launchers can report their own diagnostics before the 180-second host hook timeout
- measure elapsed time with a PowerShell monotonic `Stopwatch` and a POSIX wall-clock deadline instead of treating retry count as seconds
- clamp the optional `MONK_AGENT_READY_TIMEOUT` override to 1–150 seconds, enabling fast deterministic tests without weakening the production ceiling
- keep all three POSIX and all three PowerShell launcher copies byte-identical
- add isolated readiness-deadline regressions to the existing macOS, Ubuntu, and Windows CI jobs

## Why

Each retry includes a health request that can take up to two seconds plus a one-second sleep. The prior 180-iteration loop therefore consumed at least the entire host timeout and could take roughly 540 seconds when the local endpoint accepted connections but did not respond. The host could terminate SessionStart before users saw the launcher's actionable log paths.

The new loop checks elapsed time again before sleeping, so a live-but-unhealthy companion exits within the configured wall-clock budget and preserves time for the diagnostic to reach the host.

## Validation

- `./tests/start-monk-agent-readiness-timeout.sh` (repeated passes on macOS; 2-second deadline, at most two sleeps, under five seconds total)
- `/bin/sh -n` on the three shipped POSIX launchers and the POSIX regression
- byte parity checks across all generated POSIX and PowerShell launcher copies
- workflow YAML parse
- `git diff --check`
- Windows-native regression added to the `windows-latest` job; it builds an isolated sleeping fixture, asserts exit 1 and the 2-second diagnostic, and cleans up the child process and temporary state

Fixes #52

### DIFF
--- a/.antigravity-plugin/scripts/start-monk-agent.ps1
+++ b/.antigravity-plugin/scripts/start-monk-agent.ps1
@@ -7,6 +7,15 @@ $AuthClientId = if ($env:MONK_AGENT_AUTH_CLIENT_ID) { $env:MONK_AGENT_AUTH_CLIEN
 $AuthAudience = if ($env:MONK_AUTH_AUDIENCE) { $env:MONK_AUTH_AUDIENCE } else { "oaknode.com" }
 $AutospinUrl = if ($env:MONK_AUTOSPIN_URL) { $env:MONK_AUTOSPIN_URL } else { "wss://api.app.monk.io/autospin/" }
 
+# Host hooks terminate this launcher after 180 seconds. Reserve enough time for
+# this script to emit its own startup diagnostics instead of being killed first.
+$ReadyTimeoutSec = 150
+$RequestedReadyTimeout = 0
+if ($env:MONK_AGENT_READY_TIMEOUT -and
+    [int]::TryParse($env:MONK_AGENT_READY_TIMEOUT, [ref]$RequestedReadyTimeout)) {
+  $ReadyTimeoutSec = [Math]::Max(1, [Math]::Min(150, $RequestedReadyTimeout))
+}
+
 $ScriptDir = if ($PSScriptRoot) { $PSScriptRoot } else { Split-Path -Parent $MyInvocation.MyCommand.Path }
 $PluginRoot = if ($env:CLAUDE_PLUGIN_ROOT) { $env:CLAUDE_PLUGIN_ROOT } else { Split-Path -Parent $ScriptDir }
 $InstallDir = if ($env:MONK_AGENT_INSTALL_DIR) { $env:MONK_AGENT_INSTALL_DIR } else { Join-Path $HOME ".monk\bin" }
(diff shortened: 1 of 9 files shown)


Lachann:
Thanks for this — your wall-clock readiness deadline (`MONK_AGENT_READY_TIMEOUT`) was ported into the fix that shipped in plugin release **v0.1.50**. Closing without merging directly since `main` has since diverged (combined with related fastpath fixes in the same files), but full credit goes to this PR.

PR #128, opened 2026-07-21
fix: pass Antigravity MCP fallback inputs as arguments

### LINKED ISSUE #26: [Bug bounty] Antigravity Python fallback breaks on apostrophes in the config path

jamilahmadzai:
## Summary

- pass the Antigravity config path and server URL to Python through `sys.argv` instead of interpolating them into generated Python source
- apply the fix to all three rendered launcher copies
- add a network-free regression that exercises every rendered launcher with an apostrophe-containing config path and verifies existing JSON is preserved
- run the regression in the Ubuntu/macOS install workflow and enforce rendered-copy parity

Fixes #26

## Validation

- `./tests/start-monk-agent-antigravity-python-path.sh` (5 consecutive passes)
- `sh -n` on the changed launchers and new shell fixtures
- byte-for-byte parity checks for all three rendered launchers
- workflow YAML parse
- `git diff --check`

### DIFF
--- a/.antigravity-plugin/scripts/start-monk-agent.sh
+++ b/.antigravity-plugin/scripts/start-monk-agent.sh
@@ -86,17 +86,18 @@ register_antigravity_mcp() {
     fi
   elif command -v python3 >/dev/null 2>&1; then
     if [ "$has_existing" = "1" ]; then
-      python3 -c "
+      python3 -c '
 import json, sys
-cfg = json.load(open('$mcp_cfg'))
-servers = cfg.get('mcpServers')
+with open(sys.argv[1], encoding="utf-8") as handle:
+    cfg = json.load(handle)
+servers = cfg.get("mcpServers")
 if not isinstance(servers, dict):
     servers = {}
-cfg['mcpServers'] = servers
-servers['monk'] = {'serverUrl': '$server_url'}
+cfg["mcpServers"] = servers
+servers["monk"] = {"serverUrl": sys.argv[2]}
 json.dump(cfg, sys.stdout, indent=2)
 print()
-" >"$tmp"
+' "$mcp_cfg" "$server_url" >"$tmp"
     else
       printf '{"mcpServers":{"monk":{"serverUrl":"%s"}}}\n' "$server_url" >"$tmp"
     fi
--- a/.github/workflows/install-e2e.yml
+++ b/.github/workflows/install-e2e.yml
@@ -27,6 +27,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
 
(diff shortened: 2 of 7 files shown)


Lachann:
Thanks for the PR — we ported this approach (passing values via `sys.argv` instead of interpolating into the generated Python source) into the fix that shipped in v0.1.53, extending it to two more code paths with the same bug. This PR no longer merges cleanly against current `main`, so I'm closing it without merging, but credit for the fix stands.

PR #130, opened 2026-07-21
fix: detect Antigravity Monk MCP structurally

### LINKED ISSUE #19: [Bug bounty] unrelated "monk" text skips Antigravity MCP registration
### Stage

Connect

### Coding agent

Antigravity

### Summary

The POSIX launcher silently skips Antigravity MCP registration whenever the exact JSON string `"monk"` appears anywhere in `~/.gemini/config/mcp_config.json`, even when `.mcpServers.monk` is absent.

The guard is a whole-file text search:

```sh
grep -q '"monk"' "$mcp_cfg"
```

As a result, an unrelated preference, server description, or other JSON value containing the exact string prevents the launcher from adding the Monk server. Antigravity then has no Monk MCP connection even though `monk-agent` is healthy, and the launcher exits successfully without explaining why.

### Reproduction

1. Create `~/.gemini/config/mcp_config.json` with a healthy unrelated server and the exact string `"monk"` outside the Monk server key:

```json
{
  "preferences": {
    "notes": "monk"
  },
  "mcpServers": {
    "existing": {
      "serverUrl": "http://127.0.0.1:9000/mcp"
    }
  }
}
```

2. Run `scripts/start-monk-agent.sh` while the companion health endpoint is available.
3. Inspect the config:

```sh
jq -e '.mcpServers.monk.serverUrl == "http://127.0.0.1:7419/mcp"' \
  ~/.gemini/config/mcp_config.json
```

The assertion exits 1. The file is unchanged and the launcher prints no registration warning.

I also reproduced this deterministically with an isolated HOME, a stubbed healthy companion response, and `MONK_AGENT_PATH=/usr/bin/true`; the launcher exits 0, but the assertion above exits 1.

### Expected

The idempotence check should inspect the JSON path `.mcpServers.monk`, not search the whole file for the string `"monk"`. If that key is missing, the launcher should add it while preserving unrelated configuration.

### Actual

Any unrelated occurrence of the exact string `"monk"` causes an early return, so Antigravity remains disconnected from Monk.

### Setup

- macOS 26.5.2, arm64
- Codex CLI 0.144.5
- Monk plugin 0.1.42
- jq 1.7.1
- The same guard is present in all three shipped POSIX launcher copies

### Severity

Major: connect-stage blocker for affected Antigravity configurations. The companion is healthy and the launcher reports success, but the required MCP server is never registered.

### Suggested fix

Use a structured check such as `jq -e '.mcpServers.monk != null'` (with the existing Python fallback where appropriate), then preserve the current idempotent return only when that specific key exists. Add a regression fixture where an unrelated JSON field equals `"monk"`.

GioLealhmx:
## Summary

- Detect existing Antigravity Monk MCP registration by parsing `.mcpServers.monk` instead of grepping the whole JSON file.
- Preserve unrelated config values that happen to equal `"monk"`.
- Keep the three POSIX launcher copies in sync.
- Add a regression test covering unrelated `"monk"` text, registration, preservation of an existing server, and second-run idempotence.

## Root cause and impact

The launcher skipped Antigravity MCP registration whenever `"monk"` appeared anywhere in `mcp_config.json`. An unrelated preference or description could therefore suppress `.mcpServers.monk`, leaving Antigravity disconnected even when the companion was healthy.

## Validation

- `& 'C:\Program Files\Git\usr\bin\bash.exe' -lc 'PATH=/usr/bin:/bin:$PATH; tests/start-monk-agent-antigravity-registration.sh'`
- `& 'C:\Program Files\Git\usr\bin\bash.exe' -lc 'PATH=/usr/bin:/bin:$PATH; sh -n tests/start-monk-agent-antigravity-registration.sh; sh -n scripts/start-monk-agent.sh; sh -n plugins/monk/scripts/start-monk-agent.sh; sh -n .antigravity-plugin/scripts/start-monk-agent.sh'`
- `git diff --check`
- Root/plugin/Antigravity launcher copies are byte-identical.

Fixes #19

Prepared with Codex assistance and validated locally with Git Bash on Windows.

### DIFF
--- a/.antigravity-plugin/scripts/start-monk-agent.sh
+++ b/.antigravity-plugin/scripts/start-monk-agent.sh
@@ -64,12 +64,34 @@ health_url="http://$host:$port/.well-known/oauth-protected-resource"
 # Register monk in Antigravity's global MCP config if ~/.gemini/config/ exists.
 # Idempotent — skips if already registered. Uses jq when available; prints
 # manual instructions otherwise.
+antigravity_monk_mcp_registered() {
+  cfg_path="$1"
+  if command -v jq >/dev/null 2>&1; then
+    jq -e '(.mcpServers // {}) as $servers | ($servers | type) == "object" and ($servers | has("monk"))' "$cfg_path" >/dev/null 2>&1
+    return $?
+  fi
+  if command -v python3 >/dev/null 2>&1; then
+    python3 -c "
+import json, sys
+try:
+    with open(sys.argv[1], encoding='utf-8') as fh:
+        cfg = json.load(fh)
+except Exception:
+    sys.exit(1)
+servers = cfg.get('mcpServers') if isinstance(cfg, dict) else None
+sys.exit(0 if isinstance(servers, dict) and 'monk' in servers else 1)
+" "$cfg_path"
+    return $?
+  fi
+  return 1
+}
+
 register_antigravity_mcp() {
   gemini_cfg="$HOME/.gemini/config"
   [ -d "$gemini_cfg" ] || return 0
   mcp_cfg="$gemini_cfg/mcp_config.json"
   server_url="http://$host:$port/mcp"
-  if [ -f "$mcp_cfg" ] && grep -q '"monk"' "$mcp_cfg" 2>/dev/null; then
+  if [ -f "$mcp_cfg" ] && antigravity_monk_mcp_registered "$mcp_cfg"; then
     return 0
   fi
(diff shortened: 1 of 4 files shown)


nooga:
Closing as already fixed — commit `27ce3aa` (PR #185, "Fixes monk-io/monk-plugin#19") replaced the substring check with a structural `jq`/`python` `.mcpServers.monk` lookup, with grep only as a last resort. Current source already behaves the way this PR proposes.

Thanks for the report.

PR #132, opened 2026-07-22
fix: verify Monk health resource identity

### LINKED ISSUE #25: [Bug bounty] unrelated health resource is accepted as the Monk companion

nexicturbo:
## Summary

- parse the OAuth protected-resource document instead of accepting any response containing a `resource` key
- require the top-level resource value to equal the configured Monk MCP endpoint on POSIX and Windows
- apply the identity check to all three launcher copies and both Antigravity pre-invocation hooks
- add deterministic POSIX and Windows regressions to the install E2E workflow

## Root cause and impact

The launcher treated any successful local response containing the text `"resource"` as proof that `monk-agent` owned the configured endpoint. An unrelated MCP-style process on the same port could therefore make the startup hook exit successfully without starting Monk, leaving every Monk tool unavailable behind a misleading healthy launch.

The updated checks parse the response and require its top-level `resource` value to equal `http://<configured-host>:<configured-port>/mcp`. Responses naming another service, malformed JSON, and missing resource values all fail closed so the managed companion is started.

## Validation

- both new regressions fail on unmodified v0.1.46: the launcher accepts `http://127.0.0.1:9999/not-monk` and never creates a managed PID
- both regressions pass with this change: the POSIX fake companion and a real isolated Windows companion are started before the exact Monk resource is accepted
- `sh -n` passes for all changed POSIX launchers/hooks and the new regression
- PowerShell AST parsing passes for all changed Windows launchers/hooks and the new regression
- existing `tests/block-monk-windows.ps1` passes
- all three distributed POSIX launcher copies are byte-identical
- all three distributed PowerShell launcher copies are byte-identical
- `.github/workflows/install-e2e.yml` parses successfully
- `git diff --check` passes

Fixes #25.

### DIFF
--- a/.antigravity-plugin/hooks/ensure-monk-agent.ps1
+++ b/.antigravity-plugin/hooks/ensure-monk-agent.ps1
@@ -19,6 +19,7 @@ if (Get-Command bash -ErrorAction SilentlyContinue) { exit 0 }
 $Port = if ($env:MONK_AGENT_PORT) { $env:MONK_AGENT_PORT } else { "7419" }
 $AgentHost = if ($env:MONK_AGENT_HOST) { $env:MONK_AGENT_HOST } else { "127.0.0.1" }
 $HealthUrl = "http://${AgentHost}:$Port/.well-known/oauth-protected-resource"
+$HealthResource = "http://${AgentHost}:$Port/mcp"
 
 function Write-Json {
   param([object]$Object)
@@ -28,7 +29,8 @@ function Write-Json {
 function Test-AgentRunning {
   try {
     $Response = Invoke-WebRequest -Uri $HealthUrl -UseBasicParsing -TimeoutSec 2
-    return $Response.Content -match '"resource"'
+    $Document = $Response.Content | ConvertFrom-Json -ErrorAction Stop
+    return [string]$Document.resource -ceq $HealthResource
   } catch {
     return $false
   }
--- a/.antigravity-plugin/hooks/ensure-monk-agent.sh
+++ b/.antigravity-plugin/hooks/ensure-monk-agent.sh
@@ -12,17 +12,18 @@ set -eu
 port="${MONK_AGENT_PORT:-7419}"
 host="${MONK_AGENT_HOST:-127.0.0.1}"
 health_url="http://$host:$port/.well-known/oauth-protected-resource"
+health_resource="http://$host:$port/mcp"
 
 is_running() {
   if command -v curl >/dev/null 2>&1; then
-    curl -fsS --max-time 2 "$health_url" 2>/dev/null | grep -q '"resource"'
-    return $?
+    response="$(curl -fsS --max-time 2 "$health_url" 2>/dev/null)" || return 1
+  elif command -v wget >/dev/null 2>&1; then
+    response="$(wget -q -T 2 -O - "$health_url" 2>/dev/null)" || return 1
+  else
+    return 1
   fi
-  if command -v wget >/dev/null 2>&1; then
-    wget -q -T 2 -O - "$health_url" 2>/dev/null | grep -q '"resource"'
-    return $?
-  fi
-  return 1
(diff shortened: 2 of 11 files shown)


nexicturbo:
Implementation and cross-platform validation are complete. GitHub has disabled the fork author's Ready for review control while the workflow awaits upstream approval. Please approve the workflow and mark this PR ready when convenient.

nexicturbo:
Rebased this completed fix onto current `v0.1.49` (`f8d0170`) and reran the cross-platform validation:

- Windows health-identity regression passed against the installed `monk-agent`
- POSIX health-identity regression passed
- PowerShell AST and shell syntax checks passed
- existing Windows block-hook regression passed
- distributed launcher copies remain byte-identical

GitHub still denies the fork author permission to transition this draft. Maintainer action needed: approve the workflow and mark the PR ready for review.

Lachann:
Thanks for this — your health-resource identity check across all vulnerable call sites was ported into the fix that shipped in plugin release **v0.1.50**. Closing without merging directly since `main` has since diverged (combined with related health-check fixes in the same files), but full credit goes to this PR.

PR #138, opened 2026-07-22
Restart healthy launcher when configuration changes

### LINKED ISSUE #21: [Bug bounty] healthy Linux launcher ignores changed auth configuration
### Stage

Connect

### Coding agent

Codex

### Summary

On non-Darwin POSIX hosts, a healthy `monk-agent` is reused without checking whether the authentication or autospin configuration requested by the current launcher still matches the environment of the running process.

Changing `MONK_AUTH_URL`, `MONK_AGENT_AUTH_CLIENT_ID`, `MONK_AUTH_AUDIENCE`, or `MONK_AUTOSPIN_URL` while the companion remains healthy does not restart it. The launcher exits 0, but the process continues using the previous values.

The macOS path already avoids this problem by checking the generated launchd configuration before reusing the process. The non-Darwin fast path checks only binary-update state and health:

```sh
if [ "$os" != "Darwin" ] && [ "$agent_updated" = "0" ] && is_running; then
  ...
  exit 0
fi
```

### Deterministic reproduction

I reproduced this without network or cloud dependencies using:

- an isolated `HOME`;
- a fake managed `monk-agent` that records the four configuration variables when it starts;
- a health stub that becomes ready after the fake process starts;
- `MONK_AGENT_AUTO_UPDATE=0` so the local managed fixture is used.

Run the launcher twice against the same healthy process:

```sh
MONK_AUTH_URL=https://auth-one.invalid scripts/start-monk-agent.sh
MONK_AUTH_URL=https://auth-two.invalid scripts/start-monk-agent.sh
```

Then assert that the fixture started twice and the second start received `https://auth-two.invalid`.

### Actual

The regression exits 1 with:

```text
healthy Linux launcher ignored the changed auth configuration
```

The start log contains exactly one entry, with `https://auth-one.invalid`. The second launcher invocation exits successfully without starting a process with the new value.

### Expected

The second invocation should restart the companion exactly once with the new configuration. A third invocation with the same configuration should reuse it.

### User impact

Operators can change the documented launcher overrides and receive a successful session start while OAuth is still routed to the previous issuer/client/audience or the companion remains connected to the previous autospin endpoint. The visible environment and running process disagree, which makes the resulting authentication failure difficult to diagnose.

### Setup

- macOS 26.5.2 host with `uname` stubbed to Linux for the POSIX background-process path
- arm64
- Codex CLI 0.144.5
- Monk plugin 0.1.42
- Released `main` fast path verified by source inspection

The PowerShell launcher has the same reuse shape: it checks binary state and health, then exits before applying `$AuthUrl`, `$AuthClientId`, `$AuthAudience`, or `$AutospinUrl`. I have not claimed a native Windows reproduction.

### Severity

Major: reproducible connect-stage configuration drift. The launcher reports success while the companion keeps authentication/backend settings that no longer match the current invocation.

### Suggested fix

Persist a small launcher-configuration fingerprint alongside the PID/path metadata and require it to match before taking the healthy-process fast path. Include the authentication/autospin values and stable plugin configuration, but keep volatile shell `PATH` out of the fingerprint. Add a regression proving one restart on a real configuration change and reuse afterward.

### Tested implementation status

I implemented this locally without pushing it into the unrelated open PR:

- POSIX and PowerShell launchers persist the executable, host, port, authentication, audience, and autospin configuration used to start the companion;
- the marker is written only after the replacement companion passes its health check;
- the healthy fast path now requires the persisted configuration to match;
- all three distributed POSIX copies and all three PowerShell copies remain byte-identical.

Validation:

- the three-invocation configuration regression passes: restart once on the changed auth URL, then reuse it;
- the existing custom-path regression still passes: unchanged path reuse plus one restart on path change;
- the independent structured Antigravity registration regression still passes;
- every shipped/test POSIX script passes `sh -n`;
- `git diff --check` passes.

PowerShell was updated by source parity and copy checks only; native Windows execution remains unclaimed. The fix is intentionally uncommitted and unpushed while PR #11 occupies the fork's `main`, preserving the bounty's one-bug-per-issue review boundary.

leafwithered:
## What changed

- Persist a fingerprint of the launcher configuration after the agent starts successfully.
- Require both a healthy agent and a matching fingerprint before taking the warm-start fast path.
- Restart the managed agent when its binary path, bind address, auth endpoints, autospin URL, local mode, or plugin version changes.
- Apply the fix consistently to the root, rendered Monk plugin, and Antigravity launcher copies.
- Add Windows and POSIX regression tests and run them in the install E2E workflow.

## Why

The launchers previously treated a successful health response as sufficient proof that the running process matched the requested configuration. A healthy process started with stale auth or server settings therefore prevented the new configuration from taking effect.

Fixes #21.

## Impact

Users can change launcher configuration without manually finding and killing an otherwise healthy `monk-agent` process. Unchanged healthy launches retain the existing fast path.

## Root cause

The warm-start decision checked process health only. No durable representation of the effective configuration was recorded or compared on subsequent launches.

## Validation

- Windows configuration-change regression passed locally.
- All three PowerShell launcher copies parse successfully.
- All three POSIX launcher copies pass `sh -n`.
- Distributed launcher copies have identical hashes.
- `git diff --check` passes.
- Windows and POSIX regression tests are wired into the repository's existing install E2E workflow.

The POSIX runtime regression is left to the Linux/macOS CI jobs because WSL is not available on the local Windows test host.

### DIFF
--- a/.antigravity-plugin/scripts/start-monk-agent.ps1
+++ b/.antigravity-plugin/scripts/start-monk-agent.ps1
@@ -27,6 +27,7 @@ $RunDir = Join-Path $DataDir "run"
 $LogOut = Join-Path $LogDir "monk-agent.out.log"
 $LogErr = Join-Path $LogDir "monk-agent.err.log"
 $PidFile = Join-Path $RunDir "monk-agent.pid"
+$ConfigFile = Join-Path $RunDir "monk-agent.config"
 $HealthUrl = "http://${AgentHost}:$Port/.well-known/oauth-protected-resource"
 
 New-Item -ItemType Directory -Force -Path $LogDir, $RunDir | Out-Null
@@ -149,6 +150,29 @@ function Get-FileSha256 {
   return ([System.BitConverter]::ToString($Hash) -replace "-", "").ToLowerInvariant()
 }
 
+function Get-LauncherConfigFingerprint {
+  param([string]$AgentPath)
+
+  $Values = @(
+    $AgentPath,
+    $AgentHost,
+    $Port,
+    $AuthUrl,
+    $AuthClientId,
+    $AuthAudience,
+    $AutospinUrl,
+    $(if ($env:MONK_AGENT_LOCAL) { $env:MONK_AGENT_LOCAL } else { "" }),
+    $(if ($env:MONK_PLUGIN_VERSION) { $env:MONK_PLUGIN_VERSION } else { "" })
+  )
+  $Bytes = [Text.Encoding]::UTF8.GetBytes(($Values -join [char]0))
+  $Sha256 = [Security.Cryptography.SHA256]::Create()
+  try {
+    return ([BitConverter]::ToString($Sha256.ComputeHash($Bytes)) -replace "-", "").ToLowerInvariant()
+  } finally {
+    $Sha256.Dispose()
+  }
+}
+
 function Stop-ManagedAgent {
   if (-not (Test-Path $PidFile)) {
     return
@@ -243,8 +267,10 @@ if (-not (Test-Path $AgentPath)) {
 
 $AgentHashAfter = Get-FileSha256 $AgentPath
 $AgentUpdated = $AgentHashAfter -and ($AgentHashBefore -ne $AgentHashAfter)
+$LauncherConfigFingerprint = Get-LauncherConfigFingerprint -AgentPath $AgentPath
+$LauncherConfigMatches = (Test-Path $ConfigFile) -and ((Get-Content -Raw $ConfigFile).Trim() -eq $LauncherConfigFingerprint)
 
-if (-not $AgentUpdated -and (Test-AgentRunning)) {
+if (-not $AgentUpdated -and $LauncherConfigMatches -and (Test-AgentRunning)) {
   Show-SigninNudge
   exit 0
 }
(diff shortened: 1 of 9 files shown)


nooga:
Thanks for this — genuinely good diagnostic work. You correctly identified the exact two fields (`MONK_AGENT_LOCAL`, `MONK_PLUGIN_VERSION`) missing from the launcher's healthy-fastpath fingerprint on both POSIX and Windows, and the regression tests you wrote for both platforms are solid.

It doesn't apply cleanly against current main, though — we shipped an in-house fix for the same fast-path (agent path + auth/autospin drift, #10/#21/#52) a few days after this PR was opened, and this diff's context no longer lines up with that. It also introduces a second, parallel config-fingerprint file rather than extending the one that fix already added, which would leave two overlapping drift-detection mechanisms running per launch.

We're going to port the *idea* here (extending the existing state file with your two fields, adapting your test cases) rather than merge this as-is. Closing for now — happy to give it a fresh look if you want to rebase it against the current fast-path check instead of introducing a separate one.

leafwithered:
Thanks for the detailed review. I’d be happy to rebase this against the
current main branch and submit a focused replacement PR that extends the
existing fast-path state file with MONK_AGENT_LOCAL and
MONK_PLUGIN_VERSION, while adapting the POSIX and Windows regression tests.

Is this still useful, or is the in-house port already underway?

mg ***@***.***> 于 2026年8月3日周一 20:09写道：

PR #147, opened 2026-07-22
fix: fail when launcher mutex wait times out

### LINKED ISSUE #146: [Bug bounty] Windows launcher continues after mutex wait timeout
## Bounty eligibility

- [x] I installed and ran the Monk plugin and agent
- [x] I have starred this repository

## Stage

install

## Coding agent

OpenAI Codex desktop

## Plugin version

Repository `a4c99726e6e2db60798a49a122ac134a3ad7a620` (`v0.1.46`) on Windows 11 x64.

## Summary

The Windows launcher waits up to 190 seconds for its named single-instance mutex, but discards the Boolean result returned by `Mutex.WaitOne`. When the wait times out, the script continues into installer and process-start work without owning the mutex.

Current code in all three distributed PowerShell launcher copies:

```powershell
$LauncherMutex = New-Object System.Threading.Mutex($false, "Local\monk-agent-launcher-$Port")
try {
  [void]$LauncherMutex.WaitOne([TimeSpan]::FromSeconds(190))
} catch [System.Threading.AbandonedMutexException] {
  # continue as the new owner
}

$ManagedAgentPath = Join-Path $InstallDir "monk-agent.exe"
```

`WaitOne(timeout)` returns `$false` on timeout. It does not throw. Casting the result to void removes the only indication that the launcher failed to acquire the lock.

## Safe reproduction

I used an isolated regression that does not touch a real agent, account, WSL distro, or cloud resource:

1. Copy each shipped PowerShell launcher to a temporary directory.
2. Change only the wait duration in the temporary copy from 190 seconds to 100 milliseconds.
3. Acquire the exact named mutex from the parent test process.
4. Point `MONK_AGENT_PATH` at a temporary fake command that writes a marker if started.
5. Run the copied launcher on a random high port.

Against current main, the launcher passes the timed-out wait and reaches the fake process start. The regression fails because work occurred without ownership of the mutex.

With the tested fix, the launcher exits nonzero after the bounded wait, prints:

```text
Timed out waiting for another monk-agent launcher on port <port> to finish.
```

and never starts the fake agent.

## Expected behavior

The launcher should proceed only after `WaitOne` returns true or after an `AbandonedMutexException`, which grants ownership to the waiting thread. A normal timeout should exit nonzero with a clear message and must not enter installation, stop, or start paths.

## Actual behavior

A launcher that times out waiting for another invocation continues without lock ownership. If the first invocation is still installing or starting the agent, both launchers can mutate the managed binary, PID state, logs, and port lifecycle concurrently.

## Impact

The mutex exists specifically to serialize duplicate Windows hook invocations. A slow or stuck first invocation can outlive 190 seconds. At that point the current code disables its own race protection and starts a second launcher in the same critical section.

## Duplicate check

I searched open and closed issues and pull requests for `mutex timeout`, `WaitOne`, `launcher mutex timeout`, and `190 seconds`. I found no matching report. Issue #109 concerns the mutex being scoped by port and a shared installer race. This report is a separate failure mode where even the selected mutex is not owned after its bounded wait expires.

## Tested fix

Store the return value from `WaitOne`, set it to true for the abandoned-mutex ownership case, and fail before any installer or process work when it is false. I validated the fix with:

- regression failure against current main
- regression success against all three patched launcher copies
- PowerShell parser checks for all changed scripts
- byte-for-byte parity across all three distributed launchers
- existing Windows block-monk regression
- `git diff --check`

## Severity

major

bobthecomputer:
## Summary

- require the Windows launcher to own its named mutex before entering installer or process lifecycle work
- fail with a clear nonzero result when the bounded mutex wait expires
- preserve abandoned-mutex recovery, where the waiting thread becomes the owner
- keep all three distributed PowerShell launchers byte-identical
- add a deterministic regression for the timeout path to the Windows install workflow

## Root cause

`Mutex.WaitOne(timeout)` returns `false` on a normal timeout. The launcher cast that result to `void`, so a second invocation continued without owning the mutex after 190 seconds. This defeated the single-instance guard precisely when the first launcher was still slow or stuck.

## Validation

- new regression fails against current main because the old launcher reaches work after the mutex timeout
- `powershell.exe -NoProfile -ExecutionPolicy Bypass -File .\tests\start-monk-agent-mutex-timeout.ps1`
- PowerShell parser checks for all changed scripts
- all three launcher copies have identical SHA-256 hashes
- `powershell.exe -NoProfile -ExecutionPolicy Bypass -File .\tests\block-monk-windows.ps1`
- `git diff --check`

Fixes #146

### DIFF
--- a/.antigravity-plugin/scripts/start-monk-agent.ps1
+++ b/.antigravity-plugin/scripts/start-monk-agent.ps1
@@ -210,9 +210,14 @@ if (Test-Path $TelemetryHelper) {
 # `exit` points need no explicit unlock.
 $LauncherMutex = New-Object System.Threading.Mutex($false, "Local\monk-agent-launcher-$Port")
 try {
-  [void]$LauncherMutex.WaitOne([TimeSpan]::FromSeconds(190))
+  $LauncherMutexAcquired = $LauncherMutex.WaitOne([TimeSpan]::FromSeconds(190))
 } catch [System.Threading.AbandonedMutexException] {
   # A previous holder died mid-start; we now own the mutex and continue.
+  $LauncherMutexAcquired = $true
+}
+if (-not $LauncherMutexAcquired) {
+  Write-Error "Timed out waiting for another monk-agent launcher on port $Port to finish."
+  exit 1
 }
 
 $ManagedAgentPath = Join-Path $InstallDir "monk-agent.exe"
--- a/.github/workflows/install-e2e.yml
+++ b/.github/workflows/install-e2e.yml
@@ -66,6 +66,10 @@ jobs:
         shell: pwsh
         run: .\tests\block-monk-windows.ps1
 
+      - name: Test Windows launcher mutex timeout
+        shell: pwsh
+        run: .\tests\start-monk-agent-mutex-timeout.ps1
+
       - name: Install monk-agent
         shell: pwsh
         env:
--- a/plugins/monk/scripts/start-monk-agent.ps1
+++ b/plugins/monk/scripts/start-monk-agent.ps1
@@ -210,9 +210,14 @@ if (Test-Path $TelemetryHelper) {
 # `exit` points need no explicit unlock.
 $LauncherMutex = New-Object System.Threading.Mutex($false, "Local\monk-agent-launcher-$Port")
 try {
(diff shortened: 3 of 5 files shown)


Lachann:

PR #155, opened 2026-07-23
fix: wait for POSIX companion shutdown before restart

### LINKED ISSUE #154: [Bug bounty] POSIX restart can report success after both companions exit
### Bounty eligibility

### Stage

connect

### Coding agent

OpenAI Codex

### Coding agent version

Codex desktop; Monk plugin v0.1.47 (271cf17)

### Repro steps

The POSIX launcher stops a recorded background companion and starts its replacement without waiting for the old process to exit:

```sh
if [ -f "$pid_file" ]; then
  old_pid="$(cat "$pid_file" 2>/dev/null || true)"
  if [ -n "$old_pid" ]; then
    kill "$old_pid" >/dev/null 2>&1 || true
  fi
fi
# replacement starts immediately
```

That creates a deterministic restart race when the old companion needs time to release its listener:

1. Start an isolated old companion fixture whose health endpoint remains available for two seconds after SIGTERM.
2. Put its PID in the launcher's normal `monk-agent.pid` path.
3. Make the managed-agent ensure step install a different executable so `agent_updated=1` and the healthy fast path is bypassed.
4. Make the replacement fixture match `monk-agent`'s healthy-port-conflict behavior: if the old endpoint is still present, defer to it and exit 0.
5. Run the unmodified v0.1.47 `scripts/start-monk-agent.sh` through its Linux background-process path, with health and process behavior isolated behind local filesystem stubs.
6. After the launcher returns, wait for the old process's two-second shutdown to complete and check whether the replacement is alive.

Observed on repeated runs:

```text
launcher_exit=0
old_companion_alive=no
replacement_companion_alive=no
replacement_start_attempts=1
launcher returned success before the replacement companion was running
```

The replacement starts while the old companion still owns the simulated endpoint and exits cleanly. The launcher's first readiness check then reaches the old companion and returns success. Once the delayed old shutdown finishes, no companion remains and the PID file points to the exited replacement.

This is distinct from existing reports:

- #98 concerns signaling an unrelated process through a stale PID; this reproduction uses the correct live PID for the managed old companion.
- #52 concerns the readiness loop exceeding the host timeout; this launcher returns quickly and successfully.
- #39 concerns waiting after the newly started child has already exited; here the readiness check succeeds against the *old* child before liveness can be evaluated.
- #25/#132 concern accepting a non-Monk health responder; both fixtures model the expected Monk resource response and the bug is the restart handoff order.

Repository searches for `restart handoff`, `replacement companion`, `old agent port`, `port conflict restart`, `start_with_background_process wait`, and `binary update launcher` found no same-scope issue or PR.

### Expected behavior

A managed POSIX restart should complete the old companion's shutdown before launching its replacement. Returning success must mean the replacement process—not the process being stopped—has passed readiness and remains running.

### Actual behavior

The replacement can start before the old companion releases the port, defer to the old healthy process, and exit. The launcher then mistakes the stopping old process for successful replacement readiness and exits 0. Shortly afterward both processes are gone and Monk MCP is unavailable despite the successful SessionStart hook.

### Severity (your guess)

major

### OS

macOS 26.5.2, Apple Silicon, using the shipped POSIX launcher's Linux background-process path with deterministic process/health stubs

### monkd version

Not involved; failure occurs in local companion restart handoff before any monkd operation

### Target cloud

None; local companion only

### Integration (if relevant)

Monk plugin POSIX SessionStart launcher for Codex and Claude Code

jamilahmadzai:
## Summary

- wait up to 10 seconds for the managed POSIX companion to stop before launching its replacement
- fail safely if the old process does not stop, instead of starting a competing process
- keep the root, packaged Monk, and Antigravity launcher copies byte-identical
- add a deterministic restart-handoff regression to Linux and macOS CI

## Root cause

The POSIX background launcher sent `SIGTERM` to the recorded process and immediately started the replacement. If the old companion still owned the port, the replacement could defer to the apparently healthy old process and exit. The launcher's first readiness check could then succeed against the process being stopped, so SessionStart returned success shortly before both companions were gone.

Waiting for the managed process to exit closes that handoff window. The bounded timeout also prevents the launcher from hanging indefinitely or creating a competing replacement when shutdown fails.

## Validation

- `./tests/start-monk-agent-restart-handoff.sh` (five consecutive passes)
- `sh -n` on all three launcher copies and the new regression
- byte-for-byte comparison of all three launcher copies
- workflow YAML parse
- `git diff --check`

Fixes #154.

### DIFF
--- a/.antigravity-plugin/scripts/start-monk-agent.sh
+++ b/.antigravity-plugin/scripts/start-monk-agent.sh
@@ -464,7 +464,17 @@ start_with_background_process() {
   if [ -f "$pid_file" ]; then
     old_pid="$(cat "$pid_file" 2>/dev/null || true)"
     if pid_matches_executable "$old_pid" "$agent_path" && kill -0 "$old_pid" >/dev/null 2>&1; then
-      kill "$old_pid" >/dev/null 2>&1 || true
+      if kill "$old_pid" >/dev/null 2>&1; then
+        stop_tries=0
+        while kill -0 "$old_pid" >/dev/null 2>&1 && [ "$stop_tries" -lt 10 ]; do
+          sleep 1
+          stop_tries=$((stop_tries + 1))
+        done
+        if kill -0 "$old_pid" >/dev/null 2>&1; then
+          echo "monk-agent did not stop within 10 seconds; refusing to start a competing replacement." >&2
+          return 1
+        fi
+      fi
     fi
   fi
   export MONK_AUTH_URL="$auth_url"
--- a/.github/workflows/install-e2e.yml
+++ b/.github/workflows/install-e2e.yml
@@ -27,6 +27,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
 
+      - name: Test POSIX restart handoff
+        run: ./tests/start-monk-agent-restart-handoff.sh
+
       - name: Install monk-agent
         shell: bash
         env:
--- a/plugins/monk/scripts/start-monk-agent.sh
+++ b/plugins/monk/scripts/start-monk-agent.sh
@@ -464,7 +464,17 @@ start_with_background_process() {
   if [ -f "$pid_file" ]; then
     old_pid="$(cat "$pid_file" 2>/dev/null || true)"
     if pid_matches_executable "$old_pid" "$agent_path" && kill -0 "$old_pid" >/dev/null 2>&1; then
-      kill "$old_pid" >/dev/null 2>&1 || true
+      if kill "$old_pid" >/dev/null 2>&1; then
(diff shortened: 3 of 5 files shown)


jamilahmadzai:
Rebased onto current main/v0.1.54 and pushed head `956a038`. The refresh preserves the current executable-identity guard while retaining the bounded shutdown wait. Validation passed: restart-handoff regression 5/5, POSIX syntax checks, three-copy parity, workflow YAML parsing, and `git diff --check`.

Lachann:
Landed via our v0.1.55 work — crediting this PR directly. Closing without merging, as main has diverged since.

PR #182, opened 2026-07-30
fix: retry Linux background launcher after early exit

### LINKED ISSUE #180: [Bug bounty] Linux agent start has no port-conflict recovery unlike macOS launchd path
**Stage:** operate · **Severity:** medium (stability) · **OS:** Linux · **Agent:** all · **plugin/agent:** v0.1.52

## Summary

The `start-monk-agent.sh` script handles port conflicts differently on macOS and Linux. On macOS, the launchd plist uses `KeepAlive` with `SuccessfulExit=false`, so the agent automatically restarts if it exits due to a port conflict. On Linux, the `start_with_background_process` function has no equivalent recovery mechanism — if the agent exits because the port is taken, the script waits for the readiness timeout and then exits 1 with no retry.

## Root cause

In `scripts/start-monk-agent.sh`, the macOS path uses:
```xml
<key>KeepAlive</key>
<dict>
  <key>SuccessfulExit</key>
  <false/>
</dict>
```
This tells launchd to restart the agent whenever it exits with a non-zero exit code. The monk-agent is designed to exit 0 when it detects another healthy agent on the port (defer), and non-zero when the port is held by a foreign process (retry).

The Linux path (`start_with_background_process`) has no such mechanism:
```bash
setsid "$agent_path" serve --host "$host" --port "$port" >>"$log_file" 2>&1 </dev/null &
pid="$!"
printf '%s\n' "$pid" >"$pid_file"
```
If the agent process exits (port conflict, crash, etc.), the script enters a readiness loop that eventually times out and exits 1. There is no restart logic.

## Reproduction

1. Start monk-agent on Linux: `./start-monk-agent.sh`
2. Simulate a port conflict by binding the same port: `nc -l 127.0.0.1 3000 &`
3. Restart monk-agent: `./start-monk-agent.sh`
4. Observe: The agent starts, detects the port conflict, exits non-zero
5. The script waits for the readiness timeout (default ~30s)
6. Script exits 1: "monk-agent did not become ready"
7. No automatic retry, no port conflict recovery

On macOS, step 5-7 would NOT happen — launchd would restart the agent after a 60s throttle.

## Impact

- Linux users experience a complete failure when another process holds the agent port
- No auto-recovery unlike macOS
- User must manually resolve the port conflict and re-run the script
- Affects Claude Code, Codex, and Cursor users on Linux

## Fix

Add a retry loop in `start_with_background_process` similar to the launchd behavior:
```bash
max_retries=3
retry_count=0
while [ "$retry_count" -lt "$max_retries" ]; do
  # ... existing start logic ...
  sleep 2
  if kill -0 "$pid" 2>/dev/null; then
    break  # Agent is running
  fi
  retry_count=$((retry_count + 1))
  if [ "$retry_count" -lt "$max_retries" ]; then
    echo "Agent exited, retrying ($retry_count/$max_retries)..." >&2
    sleep 5
  fi
done
```

Or use a systemd user service on Linux with the same `Restart=on-failure` semantics.

jamilahmadzai:
## Summary

- retry non-launchd POSIX launcher starts when the child exits before readiness
- keep retries bounded and inside the existing readiness deadline
- add a Linux regression that proves the second start is used after an early child exit

## Root cause

The macOS path delegates non-zero agent exits to launchd `KeepAlive`, but the Linux/background path started the child once. If that child exited before the health endpoint became ready, the launcher observed the dead PID, stopped waiting, and returned a not-ready error without giving the agent another chance to recover.

The retry stays limited by `MONK_AGENT_BACKGROUND_RESTARTS` with a small default and the existing readiness deadline, so persistent failures still terminate cleanly.

## Validation

- `./tests/start-monk-agent-background-retry.sh`
- `./tests/start-monk-agent-readiness-timeout.sh`
- `./tests/start-monk-agent-fastpath.sh`
- `sh -n scripts/start-monk-agent.sh plugins/monk/scripts/start-monk-agent.sh .antigravity-plugin/scripts/start-monk-agent.sh tests/start-monk-agent-background-retry.sh tests/start-monk-agent-readiness-timeout.sh tests/start-monk-agent-fastpath.sh`
- generated launcher copy parity with `cmp`
- Ruby YAML parse for `.github/workflows/install-e2e.yml`
- `git diff --check`

Fixes #180.

### DIFF
--- a/.antigravity-plugin/scripts/start-monk-agent.sh
+++ b/.antigravity-plugin/scripts/start-monk-agent.sh
@@ -23,6 +23,14 @@ elif [ "$ready_timeout" -gt 150 ]; then
   ready_timeout=150
 fi
 
+background_restart_limit="${MONK_AGENT_BACKGROUND_RESTARTS:-3}"
+case "$background_restart_limit" in
+  ''|*[!0-9]*) background_restart_limit=3 ;;
+esac
+if [ "$background_restart_limit" -gt 10 ]; then
+  background_restart_limit=10
+fi
+
 # Rendered at plugin build time; carries MONK_PLUGIN_VERSION so the agent can
 # report the real plugin version in telemetry. Guarded: an older rendered
 # plugin without the file must still launch (the agent falls back to a labeled
@@ -468,6 +476,7 @@ case "$os" in
   *) start_with_background_process ;;
 esac
 
+background_restarts=0
 ready_deadline=$(( $(date +%s) + ready_timeout ))
 while [ "$(date +%s)" -lt "$ready_deadline" ]; do
   if is_running; then
@@ -480,6 +489,13 @@ while [ "$(date +%s)" -lt "$ready_deadline" ]; do
   if [ -f "$pid_file" ]; then
     _pid="$(cat "$pid_file" 2>/dev/null || true)"
     if [ -n "$_pid" ] && ! kill -0 "$_pid" 2>/dev/null; then
+      if [ "$os" != "Darwin" ] && [ "$background_restarts" -lt "$background_restart_limit" ]; then
+        background_restarts=$((background_restarts + 1))
+        echo "monk-agent exited before readiness; retrying background start ($background_restarts/$background_restart_limit)." >&2
+        start_with_background_process
+        sleep 1
+        continue
+      fi
       break
     fi
   fi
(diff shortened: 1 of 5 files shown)


jamilahmadzai:
Closing this implementation after #180 was reviewed, downgraded to a feature request, and closed. No further bounty action is needed on this PR.

PR #189, opened 2026-07-30
fix launcher telemetry JSON escaping

### LINKED ISSUE #179: [Bug bounty] Telemetry script injects unescaped special characters into JSON payload, corrupting PostHog events
**Stage:** connect · **Severity:** medium (data corruption) · **OS:** all · **Agent:** all · **plugin/agent:** v0.1.52

## Summary

The `_msan` sanitization function in `scripts/monk-launcher-telemetry.sh` does not properly escape all JSON special characters before injecting values into the hand-built JSON payload. This causes PostHog events to be silently corrupted when property values contain characters like `<`, `>`, `&`, tabs, or control characters.

## Root cause

The `_msan` function in `monk-launcher-telemetry.sh` line:
```bash
_msan() { printf '%s' "$1" | tr -d '"\\' | tr -d '\n\r'; }
```

This function only strips double-quotes (`"`) and backslashes (`\`), plus newlines. It does NOT handle other characters that are special in JSON strings:
- `<` and `>` (HTML/XML injection in property values)
- `&` (HTML entity issues)
- Tab characters (`\t`)
- Other control characters (0x00-0x1F except 0x0A and 0x0D which are handled)
- Unicode control characters

## Reproduction

1. Set `MONK_PLUGIN_VERSION='test<version>test'` as environment variable
2. Run `start-monk-agent.sh` which sources `monk-launcher-telemetry.sh`
3. The `_payload` variable becomes: `{"plugin_version":"test<version>test",...}`
4. While `<` and `>` are technically valid in JSON strings, they cause issues with:
   - XML parsers (if PostHog ever processes events through XML pipelines)
   - HTML display in PostHog UI (XSS vector in event viewer)
5. More critically, if `MONK_POSTHOG_HOST` contains a quote: `MONK_POSTHOG_HOST='https://test"test.com'`
   The payload becomes: `{"api_key":"...","event":"...","host":"https://test"test.com",...}`
   This is INVALID JSON and the entire event is dropped by PostHog.

## Impact

- PostHog events with special characters are silently corrupted or dropped
- No error logging when events fail to send
- Potential XSS vector if PostHog UI renders event properties without escaping
- Data loss in telemetry analytics

## Fix

Replace the hand-built JSON with proper escaping:
```bash
_msan() {
  printf '%s' "$1" | sed -e 's/\\/\\\\/g' -e 's/"/\\"/g' -e 's/\t/\\t/g' -e 's/\r/\\r/g' -e 's/\n/\\n/g'
}
```

Or use `jq` to build the payload properly:
```bash
printf '{}' | jq --arg key "$_ph_key" --arg cid "$_cid" \
  --arg mlc "$_mlc" --argjson first "$_first_start" \
  --argjson installed "$_agent_installed" \
  '.api_key=$key | .event="plugin_launcher_started" | .distinct_id=$cid | .launch_client=$mlc | .first_start=$first | .agent_installed=$installed'
```

jhontejada95:
## Summary

- replace the lossy `_msan` filter with a POSIX `od`/`awk` JSON string encoder
- escape quotes, backslashes, standard escapes, and every remaining `0x00`–`0x1f` control byte while preserving UTF-8
- keep the root, packaged, and Antigravity launcher scripts byte-identical
- add a regression test to the Unix CI matrix

## Root cause

The POSIX launcher builds its early PostHog payload by hand. Its sanitizer deleted quotes, backslashes, CR, and LF, but left tabs and other control bytes in JSON strings. Those bytes make the payload invalid, while deleting valid characters also silently changed telemetry values.

## Impact

Dynamic launcher properties containing quotes, slashes, whitespace controls, or UTF-8 now remain valid and round-trip without data loss, preventing affected telemetry events from being corrupted or dropped.

## Validation

- `./tests/monk-launcher-telemetry-json.sh`
- `sh -n` on all three telemetry scripts and the regression test
- `./tests/start-monk-agent-fastpath.sh`
- `./tests/start-monk-agent-readiness-timeout.sh`
- `git diff --check`

Fixes #179

### DIFF
--- a/.antigravity-plugin/scripts/monk-launcher-telemetry.sh
+++ b/.antigravity-plugin/scripts/monk-launcher-telemetry.sh
@@ -92,16 +92,35 @@ monk_emit_launcher_event() {
   _osa="$(uname -m 2>/dev/null || printf unknown)"
   _plat="$(printf '%s' "$_osn" | tr '[:upper:]' '[:lower:]')"
 
-  # Strip characters that would break the hand-built JSON.
-  _msan() { printf '%s' "$1" | tr -d '"\\' | tr -d '\n\r'; }
+  # Encode a shell string as JSON string content. LC_ALL=C makes awk's %c
+  # byte-preserving, so UTF-8 survives while every U+0000-U+001F byte that JSON
+  # forbids is escaped. POSIX od/awk keep this usable before the agent exists.
+  _mjson() {
+    printf '%s' "$1" |
+      LC_ALL=C od -An -v -t u1 2>/dev/null |
+      LC_ALL=C awk '{
+        for (i = 1; i <= NF; i++) {
+          b = $i + 0
+          if (b == 34)       printf "\\\""
+          else if (b == 92)  printf "\\\\"
+          else if (b == 8)   printf "\\b"
+          else if (b == 9)   printf "\\t"
+          else if (b == 10)  printf "\\n"
+          else if (b == 12)  printf "\\f"
+          else if (b == 13)  printf "\\r"
+          else if (b < 32)   printf "\\u%04x", b
+          else               printf "%c", b
+        }
+      }'
+  }
 
   _payload="$(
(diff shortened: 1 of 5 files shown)


nooga:
Confirmed — this fixes #179 properly. Verified by running `_mjson` directly against quote/backslash/tab/control-byte/UTF-8 inputs and round-tripping through an independent JSON parser; all three render copies stay identical, and the new test is genuinely wired into CI. Marking as the canonical fix for that cluster.

Issue #191, opened 2026-07-30
[Bug bounty] Windows PowerShell 5.1 readiness test loses child exit code


qtcqtc:
### Bounty eligibility

- [x] I have signed up at monk.io with this GitHub account
- [x] I have used the product (installed the plugin and completed the Codex MCP login)
- [ ] I have starred this repo

### Stage

install

### Coding agent

OpenAI Codex

### Coding agent version

codex-cli 0.146.0-alpha.3.1

### Repro steps

1. On Windows 11 with Windows PowerShell 5.1, install Monk plugin v0.1.52 in Codex and complete `codex mcp login monk`.
2. Clone `monk-io/monk-plugin` at the v0.1.52 source revision.
3. From Windows PowerShell 5.1, run:

   ```powershell
   & .\tests\start-monk-agent-readiness-timeout.ps1
   ```

4. The launcher behaves correctly: it exits after the configured 2-second readiness deadline and writes the expected error text.
5. The regression test nevertheless fails at line 100 because `$Launcher.ExitCode` is `$null`.

Minimal platform repro:

```powershell
$out = Join-Path $env:TEMP "child.out"
$err = Join-Path $env:TEMP "child.err"
$p = Start-Process -FilePath (Get-Process -Id $PID).Path `
  -ArgumentList @("-NoProfile", "-Command", "exit 1") `
  -PassThru -RedirectStandardOutput $out -RedirectStandardError $err
$p.WaitForExit(15000)
$p.ExitCode # null on Windows PowerShell 5.1
```

### Expected behavior

The official Windows readiness-timeout regression test should pass under the Windows PowerShell 5.1 host used by the native `.ps1` launcher. It should observe exit code `1`, confirm the `within 2s` diagnostic, and complete in less than 10 seconds.

### Actual behavior

`Start-Process` with both output streams redirected loses the exited child's `ExitCode` on Windows PowerShell 5.1. The test reports:

```text
expected launcher exit 1, got :
```

even though the launcher produced the correct bounded-timeout diagnostic. This makes the official Windows regression test fail on a supported native shell and can hide real launcher regressions behind a harness-only failure.

I have a tested fix that uses `System.Diagnostics.ProcessStartInfo` for redirected output and stops the inherited-handle fake companion before collecting async output. After the fix:

```text
Windows block-monk fallback tests passed.
readiness_timeout_status=pass elapsed=4.54s
```

### Severity

minor

### OS

[...]

PR #194, opened 2026-07-30
fix: restart Windows agent on plugin version drift

### LINKED ISSUE #193: [Bug bounty] Windows healthy fast path ignores plugin-version changes
### Bounty eligibility

- [x] I have used the product (installed plugin v0.1.52 in Codex and completed MCP authentication)

### Stage

connect

### Coding agent

OpenAI Codex

### Coding agent version

codex-cli 0.146.0-alpha.3.1; Monk plugin v0.1.52

### Repro steps

The native Windows launcher loads the rendered `plugin-version.ps1` and passes `MONK_PLUGIN_VERSION` to a newly started companion. However, the healthy-process state fingerprint does not persist or compare that value.

I reproduced this dynamically with the released v0.1.52 PowerShell launcher:

1. Create an isolated `MONK_AGENT_HOME` and a temporary copy of `scripts/start-monk-agent.ps1`.
2. Compile a small fake `monk-agent.exe` that:
   - appends its PID and inherited `MONK_PLUGIN_VERSION` to a capture file;
   - serves the exact protected-resource JSON expected by `Test-AgentRunning`;
   - stays running until the launcher restarts it.
3. Write `plugin-version.ps1` with `old-version` and invoke the launcher. The fake companion starts and records `old-version`.
4. Invoke the launcher again unchanged. It correctly reuses the healthy process.
5. Rewrite only `plugin-version.ps1` to `new-version` and invoke the launcher again.

Released-code result:

```json
{
  "LaunchCount": 1,
  "Launches": "19364|old-version",
  "State": "agent_path=...\\fake-agent.exe\nauth_url=...\nauth_client_id=...\nauth_audience=...\nautospin_url=..."
}
```

The second plugin version is never inherited because the first companion is reused, and the state file has no `plugin_version` entry.

### Expected behavior

When a new plugin release changes `MONK_PLUGIN_VERSION`, the Windows launcher should restart a healthy companion once so the running process inherits the current plugin version. A subsequent launch with the same version should reuse it.

### Actual behavior

The Windows fast path compares only agent path and auth/autospin settings. If the managed Agent binary is unchanged across plugin releases, `$AgentUpdated` is false and the old companion is reused indefinitely. Diagnostics and telemetry therefore attribute native Windows sessions to the plugin version that originally started the process.

### Severity

minor

The companion remains reachable, but release attribution is stale and version-specific field debugging becomes unreliable until the user manually restarts the companion.

### OS

Windows 11 Pro 10.0.26200 (build 26200), Windows PowerShell 5.1.26100.8875

### monkd version

Not involved; the defect is in the native Windows companion reuse decision before monkd interaction.

### Target cloud

None; local companion lifecycle.

### Integration

OpenAI Codex / native Windows PowerShell launcher

### Duplicate check

- #183 covers the same omitted value in the Linux/non-Darwin POSIX state fingerprint. Its tested fix changes only the POSIX launcher copies.
- #110 covers the older defect where Windows did not load the rendered plugin version on a fresh start. That initial propagation is fixed in v0.1.52; this report covers reuse after the rendered version changes.

No existing report covers the current Windows healthy-process fingerprint.

### Tested fix

A focused patch stores and compares `plugin_version` in all three generated PowerShell launcher copies. The new regression test verifies:

- first launch inherits `old-version`;
- unchanged version reuses the companion;
- changing to `new-version` restarts once;
- state persists `plugin_version=new-version`;
- all shipped PowerShell launcher copies stay byte-identical.

Result:

```text
plugin_version_restart_status=pass launches=2
Windows block-monk fallback tests passed.
```

The unrelated current readiness-timeout harness failure is tracked separately in #191.

qtcqtc:
Closes #193.

The Windows launcher already loads `plugin-version.ps1`, but its healthy-process fingerprint omits the resulting `MONK_PLUGIN_VERSION`. When a plugin release reuses the same Agent binary, the existing companion therefore stays alive with stale release attribution.

This change:

- snapshots the rendered plugin version after loading `plugin-version.ps1`;
- persists it in the Windows launcher state file;
- includes it in the exact fast-path comparison;
- updates every generated PowerShell launcher copy;
- adds an isolated Windows regression with a fake healthy companion.

The regression proves all three lifecycle cases:

1. the first process inherits `old-version`;
2. an unchanged second launch reuses it;
3. changing only the rendered version to `new-version` restarts once and updates state.

Verified on Windows 11 / Windows PowerShell 5.1:

```text
plugin_version_restart_status=pass launches=2
Windows block-monk fallback tests passed.
```

All generated PowerShell launcher copies are byte-identical and `git diff --check` passes.

The unrelated readiness-timeout test harness failure on Windows PowerShell 5.1 is tracked in #191 / #192.

### DIFF
--- a/.antigravity-plugin/scripts/start-monk-agent.ps1
+++ b/.antigravity-plugin/scripts/start-monk-agent.ps1
@@ -30,6 +30,7 @@ $MonkHome = if ($env:MONK_AGENT_HOME) { $env:MONK_AGENT_HOME } else { Join-Path 
 # PowerShell counterpart of the plugin-version.sh source in start-monk-agent.sh.
 $PluginVersionScript = Join-Path $ScriptDir "plugin-version.ps1"
 if (Test-Path $PluginVersionScript) { . $PluginVersionScript }
+$PluginVersion = if ($env:MONK_PLUGIN_VERSION) { $env:MONK_PLUGIN_VERSION } else { "" }
 
 $DataDir = Join-Path $MonkHome "agent\launcher"
 $LogDir = Join-Path $DataDir "logs"
@@ -340,7 +341,8 @@ function Test-BackgroundStateConfigured {
     "auth_url=$AuthUrl",
     "auth_client_id=$AuthClientId",
     "auth_audience=$AuthAudience",
-    "autospin_url=$AutospinUrl"
+    "autospin_url=$AutospinUrl",
+    "plugin_version=$PluginVersion"
   )
   $Lines = $State -split "`r?`n"
   foreach ($Line in $Expected) {
@@ -382,7 +384,8 @@ $Process.Id | Set-Content -NoNewline $PidFile
   "auth_url=$AuthUrl",
   "auth_client_id=$AuthClientId",
   "auth_audience=$AuthAudience",
-  "autospin_url=$AutospinUrl"
+  "autospin_url=$AutospinUrl",
+  "plugin_version=$PluginVersion"
 ) -join "`n" | Set-Content -NoNewline $StateFile
 
 $ReadyTimer = [System.Diagnostics.Stopwatch]::StartNew()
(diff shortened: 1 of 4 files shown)


nooga:
Thanks for the PR — the fix itself is correct and complete across all three render copies (read and write sides of the fingerprint stay symmetric, matching the existing `autospin_url` pattern). The gap is that the new regression test isn't wired into `install-e2e.yml`, so it provides no durable protection going forward — every sibling fix in this area (including #138, which covers this same field plus more) wires its tests into that workflow. Closing as inadequate as submitted; we'll fold the equivalent fix into a combined patch for this fast-path fingerprint gap.

PR #196, opened 2026-07-30
fix: use basic parsing for Windows agent downloads

### LINKED ISSUE #195: [Bug bounty] Windows bootstrap relies on unavailable IE parser in PowerShell 5.1
### Bounty eligibility

- [x] I have used the product (installed the plugin and completed the Codex MCP login)

### Stage

install

### Coding agent

OpenAI Codex

### Coding agent version

codex-cli 0.146.0-alpha.3.1

### Repro steps

1. Use a Windows host where the Monk plugin launcher runs under Windows PowerShell 5.1, as pinned by `run-powershell.cmd`.
2. Use a Server Core image, or a fresh desktop profile where the Internet Explorer engine is unavailable or has not completed first-run setup.
3. Start from plugin v0.1.52 (`6bdde78`) without an installed `monk-agent.exe`.
4. Run:

   ```powershell
   powershell.exe -NoLogo -NoProfile -ExecutionPolicy Bypass `
     -File .\scripts\ensure-monk-agent.ps1
   ```

5. Both download calls use `Invoke-WebRequest -OutFile` without `-UseBasicParsing`.

I reproduced the platform behavior deterministically with a regression harness that shadows `Invoke-WebRequest` with the documented headless-PowerShell failure and allows the request only when `-UseBasicParsing` is present. Against the unmodified v0.1.52 installer it fails immediately:

```text
The response content cannot be parsed because the Internet Explorer engine is not available.
```

### Expected behavior

The checksum and agent archive downloads should work under the stock Windows PowerShell 5.1 host without requiring Internet Explorer components, first-run setup, or an interactive desktop profile. The installer should verify the archive, extract `monk-agent.exe`, and return its installed path.

### Actual behavior

Windows PowerShell 5.1 delegates web-response parsing to the Internet Explorer engine unless `-UseBasicParsing` is explicit. The installer therefore fails on headless/Server Core systems and affected fresh profiles before it can download the checksum or agent binary. `monk-agent` is never installed, so the Monk MCP tools cannot become available.

The launcher health probes already pass `-UseBasicParsing`; the two file downloads in all three shipped `ensure-monk-agent.ps1` copies are the remaining gap.

I have a tested fix that adds `-UseBasicParsing` to both downloads and a PowerShell regression test that exercises the full checksum-download, archive-download, hash-verification, extraction, and install path. After the fix:

```text
ensure_basic_parsing_status=pass downloads=2
Windows block-monk fallback tests passed.
```

### Severity (your guess)

major

### OS

Windows 11 Pro 10.0.26200, Windows PowerShell 5.1.26100.8875; also affects Windows Server Core / headless Windows PowerShell 5.1 environments

### monkd version

Not installed: the failure occurs during the companion bootstrap before `monk-agent` can expose the Monk installation flow.

### Target cloud

Local Windows bootstrap; no cloud provider selected.

### Integration (if relevant)

OpenAI Codex + native Windows PowerShell 5.1

qtcqtc:
Closes #195.

Windows PowerShell 5.1 delegates `Invoke-WebRequest` response parsing to the Internet Explorer engine unless `-UseBasicParsing` is explicit. On Server Core and fresh profiles where that engine is unavailable or uninitialized, the native Windows bootstrap fails before it can download either the checksum or `monk-agent.exe`.

This change:

- adds `-UseBasicParsing` to the checksum and archive downloads in all three shipped PowerShell installer copies;
- adds a regression harness that simulates the headless PowerShell 5.1 failure;
- exercises the full mocked download, checksum verification, ZIP extraction, install, and output-path flow;
- verifies the generated PowerShell installer copies remain byte-identical.

The regression test fails against v0.1.52 with:

```text
The response content cannot be parsed because the Internet Explorer engine is not available.
```

After the fix, both Windows PowerShell 5.1 and the current host pass:

```text
ensure_basic_parsing_status=pass downloads=2
Windows block-monk fallback tests passed.
```

`git diff --check` also passes.

### DIFF
--- a/.antigravity-plugin/scripts/ensure-monk-agent.ps1
+++ b/.antigravity-plugin/scripts/ensure-monk-agent.ps1
@@ -111,7 +111,11 @@ $ChecksumTmp = Join-Path $InstallDir ".monk-agent.tmp.sha256"
 $ExtractDir = Join-Path $InstallDir ".monk-agent.extract"
 New-Item -ItemType Directory -Force -Path $InstallDir | Out-Null
 
-Invoke-WebRequest -Uri $ChecksumUrl -OutFile $ChecksumTmp
+# Windows PowerShell 5.1 otherwise delegates response parsing to the Internet
+# Explorer engine, which is unavailable on Server Core and can be uninitialized
+# on fresh desktop profiles. Downloads are files, so always use the independent
+# basic parser.
+Invoke-WebRequest -Uri $ChecksumUrl -OutFile $ChecksumTmp -UseBasicParsing
 
 $Expected = ((Get-Content -Raw $ChecksumTmp).Trim() -split "\s+")[0].ToLowerInvariant()
 
@@ -125,7 +129,7 @@ if ((Test-Path $Target) -and (Get-Item $Target).Length -gt 0 -and (Test-Path $Ch
 }
 
 Write-Host "Installing monk-agent from $Url"
-Invoke-WebRequest -Uri $Url -OutFile $ArchiveTmp
+Invoke-WebRequest -Uri $Url -OutFile $ArchiveTmp -UseBasicParsing
 
 $Actual = Get-FileSha256 $ArchiveTmp
 if ($Actual -ne $Expected) {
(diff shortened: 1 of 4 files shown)


nooga:
Thanks for the PR — the fix itself is correct and complete across all three render copies, and the regression test is genuinely thorough (mocks the exact IE-engine failure, verifies byte-identity across copies). The gap is the same one as #194: the new test isn't wired into `install-e2e.yml`, so it provides no durable protection going forward — every peer fix in this area (#138, #155) adds that CI step. Closing as inadequate for merge as submitted; would take another look at a version with the workflow wiring added.

PR #209, opened 2026-07-31
fix: block-monk guard fails closed when helper errors (#100)

### LINKED ISSUE #100: [Bug bounty] Windows block-monk guard fails open when helper errors

leanworld7-netizen:
## Summary

When the `monk-agent` binary exists but errors (crash, bad output, non-zero exit), the `block-monk` hook previously exited with the binary's error code. The host interprets any non-deny result as "allow", letting `monk` shell commands through — the guard **fails open** instead of closed.

## Fix

If the binary returns a non-zero exit code, fall through to the native fallback path (PowerShell regex / POSIX grep) instead of propagating the error. The fallback is biased toward **BLOCKING**, so a helper error now results in a safe deny rather than an allow.

Applied to both `block-monk.ps1` (Windows) and `block-monk.sh` (POSIX), across all three copy locations (`hooks/`, `plugins/monk/hooks/`, `.antigravity-plugin/hooks/`).

Fixes #100

Bounty eligibility: I have signed up at monk.io and used the product.

### DIFF
--- a/hooks/block-monk.ps1
+++ b/hooks/block-monk.ps1
@@ -16,7 +16,10 @@ $agent = if ($env:MONK_AGENT_PATH) { $env:MONK_AGENT_PATH } else { Join-Path $ag
 # reads raw UTF-8 (BOM-tolerant) from the inherited stdin and is unaffected.
 if (Test-Path $agent) {
   & $agent hook block-monk --format claude
-  exit $LASTEXITCODE
+  if ($LASTEXITCODE -eq 0) { exit 0 }
+  # Binary errored — fall through to fallback instead of failing open (ENG-100).
+  # A non-zero exit from the helper previously propagated to the host, which
+  # interprets any non-deny result as "allow", letting `monk` commands through.
 }
 
 # Fallback: binary unavailable. Read stdin as UTF-8 (BOM stripped) so the payload
--- a/hooks/block-monk.sh
+++ b/hooks/block-monk.sh
@@ -25,6 +25,9 @@ if [ -x "$agent" ]; then
   if printf '%s' "$input" | "$agent" hook block-monk --format claude; then
     exit 0
   fi
+  # Binary errored — fall through to fallback instead of failing open (ENG-100).
+  # A non-zero exit from the helper previously propagated to the host, which
+  # interprets any non-deny result as "allow", letting `monk` commands through.
 fi
 
 # Fallback: binary unavailable. No jq here, so pull just the `command` string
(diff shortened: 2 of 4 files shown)


nooga:
Closing as superseded — commit `904dd08` (ENG-425) already made both hooks fall through to the safe fallback on any helper failure, and the `.ps1` now validates the JSON decision shape rather than just the exit code. This PR's diff is against the pre-fix code.

Thanks for the report.

Issue #215, opened 2026-07-31
[Bug bounty] block-monk fallback regex bypass via eval/xargs command wrappers


leanworld7-netizen:
## Bounty eligibility

- [x] I have signed up at monk.io with this GitHub account
- [x] I have used the product (installed the plugin and ran it, rather than only reading the code)
- [x] I have starred this repository

## Stage

serve

## Summary

The block-monk.sh fallback regex (used when the monk-agent binary is unavailable) can be bypassed using command wrappers not in its allowlist, such as eval, xargs, or find -exec. This allows shell-out to the monk CLI when the binary is missing, defeating the guard.

The wrapper list in the regex only covers: sudo, command, env, exec, nohup, time, nice. But eval, xargs, bash -c, sh -c, find -exec, awk system(), perl system(), python os.system() are NOT covered. A command like eval monk deploy would pass through the fallback filter.

## Repro steps

1. Remove or rename the monk-agent binary so the fallback path activates.
2. In Claude Code, issue a Bash command: eval "monk deploy"
3. The block-monk hook fallback regex does not match because eval is not in the wrapper list and monk does not appear in command position after a separator or at the start of the string.
4. The command passes through to execution.

### Analysis

The fallback regex in hooks/block-monk.sh:

(^|[;&|(){}])[[:space:]]*(sudo|command|env|exec|nohup|time|nice)?[[:space:]]*([^[:space:];&|(){}]*[/])?monkd?(.(exe|cmd|bat|ps1))?([[:space:]]|$)

For the command eval monk deploy after quote/backslash stripping:

- Position 0: ^ matches, eval is not in the wrapper list, path group does not match (no trailing /), monkd? tries to match eval which fails.
- No separator character appears before monk in the string.
- The regex does not match anywhere, so the command is allowed.

Same bypass applies to: xargs monk, bash -c monk, find . -exec monk, sh -c monk.

## Expected behavior

The fallback should block all monk CLI invocations regardless of the wrapper command used. The design comment says False positives only ever BLOCK, never allow but this is a false negative (allows monk through).

## Actual behavior

eval monk deploy, xargs monk, and similar commands bypass the fallback regex and are allowed to execute.

## Severity (my guess)

minor: only affects the fallback path when the monk-agent binary is missing. The primary binary-based path uses a proper tokenizer and is not affected.

## Suggested fix

[...]

PR #217, opened 2026-07-31
fix: uninstaller fails closed + block-monk regex wrappers (#214, #215)

### LINKED ISSUE #214: [Bug bounty] POSIX uninstaller kills any process when process name cannot be determined
## Bounty eligibility

- [x] I have starred this repository

## Stage

install

## Summary

The POSIX uninstaller stop_agent() function in scripts/uninstall-monk-agent.sh kills any process identified by the PID file when the process name cannot be determined. When both /proc/PID/comm is unreadable and ps -p PID -o comm= returns empty, the script treats the empty result as a match and kills the process.

This is the POSIX sibling of #103 (Windows PID guard trusts executable basename). The Windows version was fixed by comparing full paths, but the POSIX version has a different root cause: it fails open (kills) when the process name cannot be determined, rather than failing closed (skipping).

## Repro steps

1. On Linux, start monk-agent via the plugin and note the PID in the PID file.
2. Replace that PID in the file with the PID of an unrelated long-running process.
3. Run scripts/uninstall-monk-agent.sh --yes --keep-data.

### Analysis (from reading the source at scripts/uninstall-monk-agent.sh)

The relevant code path in stop_agent():

proc_name=""
if [ -f "/proc/$pid/comm" ]; then
  proc_name="$(tr -d newline < /proc/$pid/comm 2>/dev/null || true)"
fi
if [ -z "$proc_name" ] && command -v ps >/dev/null 2>&1; then
  proc_name="$(ps -p $pid -o comm= 2>/dev/null || true)"
fi
if [ -z "$proc_name" ] || [ "$proc_name" = "monk-agent" ]; then
  kill "$pid" >/dev/null 2>&1 || true
fi

When both /proc/PID/comm and ps -p PID -o comm= return empty (container without /proc, restricted ps, race between kill -0 check and name lookup), the condition [ -z proc_name ] is true, so the script kills the process regardless of whether it is actually monk-agent.

## Expected behavior

When the process name cannot be determined, the script should fail closed: discard the stale PID file without killing the process.

## Actual behavior

An unreadable process name is treated as a confirmed match. The script kills whatever process holds that PID.

## Severity (my guess)

major: a stale PID file can cause the uninstaller to kill an unrelated process on systems where process name lookup fails.

## Suggested fix

Change the kill condition to fail closed:

if [ -n "$proc_name" ] && [ "$proc_name" = "monk-agent" ]; then
  kill "$pid" >/dev/null 2>&1 || true
fi

Only kill when identity is positively confirmed; skip when unknown.

## Duplicate check

Searched for stop_agent, PID, uninstall, kill, proc_name, comm, and fail open. #98 concerns a stale POSIX PID terminating an unrelated process via ps basename match, but its root cause is trusting the basename monk-agent rather than the empty-name fail-open path. #103 is the Windows sibling. Neither covers the POSIX empty-process-name case.

leanworld7-netizen:
### #214: Uninstaller fails closed when process name unknown

**Before:** `if [ -z $proc_name ] || [ $proc_name = monk-agent ]` — kills when name is empty (fails open)
**After:** `if [ -n $proc_name ] && [ $proc_name = monk-agent ]` — kills only when name is CONFIRMED (fails closed)

Applied to all 3 copies: scripts/, .antigravity-plugin/scripts/, plugins/monk/scripts/.

### #215: Block-monk regex catches eval/xargs/bash/sh

Added eval, xargs, bash, sh, find, awk, perl, python to the wrapper allowlist in all 3 copies of block-monk.sh.

Both fixes are minimal and surgical.

### DIFF
--- a/.antigravity-plugin/hooks/block-monk.sh
+++ b/.antigravity-plugin/hooks/block-monk.sh
@@ -52,7 +52,7 @@ raw_command="$(printf '%s' "$input" |
 # to a real newline first (same for \r).
 raw_command="$(printf '%s' "$raw_command" | awk '{gsub(/\\n/, "\n"); gsub(/\\r/, "\r"); print}')"
 normalized="$(printf '%s' "$raw_command" | tr -d '\\' | tr -d '"' | tr -d "'")"
-if printf '%s' "$normalized" | grep -Eq '(^|[;&|`(){}])[[:space:]]*(sudo|command|env|exec|nohup|time|nice)?[[:space:]]*([^[:space:];&|`(){}]*[/\\])?monkd?(\.(exe|cmd|bat|ps1))?([[:space:]]|$)'; then
+if printf '%s' "$normalized" | grep -Eq '(^|[;&|`(){}])[[:space:]]*(sudo|command|env|exec|nohup|time|nice|eval|xargs|bash|sh|find|awk|perl|python)?[[:space:]]*([^[:space:];&|`(){}]*[/\\])?monkd?(\.(exe|cmd|bat|ps1))?([[:space:]]|$)'; then
   cat <<'JSON'
 {
   "decision": "deny",
--- a/.antigravity-plugin/scripts/uninstall-monk-agent.sh
+++ b/.antigravity-plugin/scripts/uninstall-monk-agent.sh
@@ -106,7 +106,7 @@ stop_agent() {
           if [ -z "$proc_name" ] && command -v ps >/dev/null 2>&1; then
             proc_name="$(ps -p "$pid" -o comm= 2>/dev/null || true)"
           fi
-          if [ -z "$proc_name" ] || [ "$proc_name" = "monk-agent" ]; then
+          if [ -n "$proc_name" ] && [ "$proc_name" = "monk-agent" ]; then
             kill "$pid" >/dev/null 2>&1 || true
           fi
         fi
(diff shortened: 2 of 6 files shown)


nooga:
Split review on this one, since it covers two different files/issues:

**#214 (uninstaller fail-open kill) — passes.** The `[ -n "$proc_name" ] && [ "$proc_name" = "monk-agent" ]` fail-closed change is correct and complete across all rendered copies. We'll be porting this as the canonical fix.

**#215 (block-monk fallback wrapper bypass) — doesn't fully close it.** I ran the pre- and post-patch regex against the actual reported bypasses: adding `eval|xargs|bash|sh|find|awk|perl|python` to the wrapper keyword list blocks the bare form (`bash monk`, `eval "monk deploy"`), but `bash -c 'monk deploy'` and `sh -c 'monk deploy'` — the shape #215 specifically calls out as the realistic exploit path — still get through on all three copies, since the `-c` flag breaks the adjacency the wrapper-then-`monk` pattern requires. Leaving this open rather than closing it since the uninstaller half is solid; we'll fold a fix for the remaining `-c`/quoted-argument gap into the eventual patch, and can credit an updated version of this PR if you want to take another pass at it.

Lachann:
Thanks for the report — the uninstaller half of this shipped in v0.1.54, though via broader launcher hardening (monk-agent#218) rather than a direct port of this PR. This PR no longer merges cleanly against current `main`, so I'm closing it without merging, but credit for identifying the fail-open bug stands.

PR #263, opened 2026-08-05
fix: keep native Antigravity hook for unusable Bash

### LINKED ISSUE #258: [Bug bounty] unusable Windows bash stub suppresses Antigravity ensure hook
> **Current release update (Aug. 6):** Reproduced unchanged on the current v0.1.54 public artifact `2a4f9cf`. PR #263 carries the minimal fix and an isolated regression covering both unusable and usable Bash commands.

### Stage

connect

### Coding agent

OpenAI Codex

### Coding agent version

### Repro steps

The native Windows Antigravity `ensure-monk-agent.ps1` hook silently delegates to its POSIX sibling whenever `Get-Command bash` returns any command:

```powershell
if (Get-Command bash -ErrorAction SilentlyContinue) { exit 0 }
```

Windows can expose the legacy WSL `bash.exe` launcher even when it cannot start a shell. This is the current state on my Windows 11 host:

```text
Name   : bash.exe
Source : C:\WINDOWS\system32\bash.exe

bash --version
WSL ERROR: execvpe(/bin/bash) failed: No such file or directory
exit 1
```

I reproduced the shipped hook with an isolated missing `MONK_AGENT_PATH` and unused loopback port:

```text
stdout_length=0
exit=0
```

The hook exits before its health check, binary check, or install-guidance JSON. I then added `tests/antigravity-ensure-unusable-bash-stub.ps1`, which puts a failing `bash.cmd` ahead of PATH, runs the exact shipped hook in a child Windows PowerShell process, and requires the normal `monk-agent is not installed` `injectSteps` payload. No real agent, network, OAuth state, or user data is involved.

Pristine v0.1.54 fails:

```text
FAIL: unusable bash command made the native Antigravity hook exit without its install guidance; stdout=''
```

### Expected behavior

The PowerShell hook should bow out only when the detected Bash command can actually start a shell. If the Windows/WSL launcher exists but is unusable, the native hook should continue and emit its normal running, start, or install-guidance JSON.

### Actual behavior

Command discovery is treated as proof that the POSIX sibling can run. An unusable Windows `bash.exe` makes the native hook exit successfully with no stdout, so a cold invocation receives no guidance and the companion is not started.

I validated a minimal fix in PR #263: invoke the discovered Bash command with `-lc "exit 0"` while suppressing its output, and delegate only when it exits zero. The focused regression then passes and the updated PowerShell file parses with zero errors:

```text
PASS: unusable bash command does not suppress the native Antigravity ensure hook
powershell_parse_errors=0
```

### Collision check

I searched all repository issues (open and closed, limit 500) for `Get-Command bash`, `bash.exe`, WSL/no-distro, `/bin/bash`, unusable Bash, native-hook skip, and Antigravity PowerShell delegation.

#135 is the closest report but is not this failure. It covered the security-oriented `block-monk` hook selecting a runnable Bash whose `.sh` file then failed on CRLF line endings; it was closed after v0.1.52 forced LF endings. This report covers the non-security `ensure-monk-agent` lifecycle hook when the Bash command itself cannot start any shell, reproduces on current v0.1.54, and occurs before line endings or POSIX script contents are involved. #135's suggested improvements mentioned the presence-only weakness, but its shipped `.gitattributes` fix did not address this executable-liveness case.

No issue reports the current Windows/WSL stub condition suppressing Antigravity install/start guidance.

### Severity (your guess)

major

The failure is silent and repeats on every cold model invocation: Antigravity receives neither a companion start nor the hook's actionable install guidance, while the hook exits successfully and therefore looks healthy to the host.

### OS

Windows 11; system `bash.exe` present through WSL, but `/bin/bash` unavailable. Also reproduced with an isolated failing Bash-command fixture.

### monkd version

Not involved; the defect occurs before companion/runtime access.

### Target cloud

None; local Antigravity hook only.

### Integration

Monk plugin for Antigravity on Windows

aikinley:
## Summary

- probe a discovered Bash application before the native Antigravity hook yields to its POSIX sibling
- bound that liveness probe to two seconds so a hung WSL launcher cannot consume the host hook budget
- keep the PowerShell fallback active when Windows exposes a broken or hung WSL `bash.exe`
- preserve delegation when Bash really can start a shell
- add an isolated Windows regression covering failed, successful, and hung Bash launchers

On unmodified v0.1.54 the regression exits with:

```
FAIL: unusable bash command made the native Antigravity hook exit without its install guidance; stdout=''
```

With this change failed and hung launchers produce the intended `monk-agent is not installed` JSON, while a usable Bash command still suppresses duplicate native output. The hung-launcher case completes within the bounded fallback assertion.

## Validation

- `powershell -File tests/antigravity-ensure-unusable-bash-stub.ps1`
- PowerShell parser: zero errors

## Generated source note

`GENERATED.md` identifies this public tree as generated from `monk-agent/plugin`. That source repository is not publicly accessible, so this PR patches and tests the currently shipped artifact; the same production hunk and regression should be ported to the generator before the next release regeneration.

Fixes #258

### DIFF
--- a/.antigravity-plugin/hooks/ensure-monk-agent.ps1
+++ b/.antigravity-plugin/hooks/ensure-monk-agent.ps1
@@ -20,12 +20,30 @@ $ErrorActionPreference = "SilentlyContinue"
 # invocation.
 $BashUsable = $false
 $BashCommand = Get-Command bash -ErrorAction SilentlyContinue
-if ($BashCommand) {
+if ($BashCommand -and $BashCommand.CommandType -eq "Application") {
+  $BashProbe = New-Object System.Diagnostics.Process
   try {
-    & $BashCommand.Source -lc "exit 0" *> $null
-    $BashUsable = $LASTEXITCODE -eq 0
+    $BashProbe.StartInfo.FileName = $BashCommand.Source
+    $BashProbe.StartInfo.Arguments = '-lc "exit 0"'
+    $BashProbe.StartInfo.UseShellExecute = $false
+    $BashProbe.StartInfo.RedirectStandardOutput = $true
+    $BashProbe.StartInfo.RedirectStandardError = $true
+    $BashProbe.StartInfo.CreateNoWindow = $true
+    if ($BashProbe.Start()) {
+      # Drain redirected streams asynchronously so a broken launcher cannot
+      # block while reporting its own startup failure.
+      $null = $BashProbe.StandardOutput.ReadToEndAsync()
+      $null = $BashProbe.StandardError.ReadToEndAsync()
+      if ($BashProbe.WaitForExit(2000)) {
+        $BashUsable = $BashProbe.ExitCode -eq 0
+      } else {
+        $BashProbe.Kill()
+      }
+    }
   } catch {
     $BashUsable = $false
+  } finally {
+    $BashProbe.Dispose()
   }
 }
 if ($BashUsable) { exit 0 }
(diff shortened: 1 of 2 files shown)


Lachann:

PR #265, opened 2026-08-05
fix: preserve non-object Antigravity configs

### LINKED ISSUE #255: [Bug bounty] valid non-object Antigravity config aborts a healthy launcher
### Stage

connect

### Coding agent

OpenAI Codex

### Coding agent version

### Repro steps

1. Check out v0.1.54 on a POSIX shell with `python3` available and `jq` absent, so Antigravity registration takes its documented Python fallback.
2. Create `~/.gemini/config/mcp_config.json` containing syntactically valid JSON whose root is not an object:

   ```json
   [{"preserved":"still here"}]
   ```

3. Run `scripts/start-monk-agent.sh` with an otherwise healthy companion response.

I reproduced this with a network-free fixture that isolates `HOME` and `PATH`, returns the expected Monk health resource, and uses an inert local executable as the companion. The v0.1.54 launcher exits 1:

```text
Traceback (most recent call last):
  File "<string>", line 5, in <module>
AttributeError: 'list' object has no attribute 'get'
```

The same config shape reaches an assignment failure through the jq branch because `.mcpServers.monk = ...` cannot index an array root by the string `mcpServers`.

### Expected behavior

Automatic Antigravity registration is optional. A config whose JSON root has an incompatible shape should be preserved, produce a warning/manual-registration hint, and allow the already-healthy Monk launcher to succeed.

### Actual behavior

The validation guard checks only JSON syntax:

```sh
jq empty "$mcp_cfg"
# or json.load(...)
```

A top-level array therefore passes validation. The Python fallback then calls `cfg.get("mcpServers")`, raises `AttributeError`, and aborts the entire launcher under `set -e`. The optional Antigravity integration can consequently break Monk startup in Codex or Claude Code after the companion health check has already passed.

I validated a minimal fix locally: require the parsed root to be an object in both validation branches (`jq -e 'type == "object"'` / `isinstance(cfg, dict)`). The isolated regression then warns, preserves the config byte-for-byte, and passes; `sh -n scripts/start-monk-agent.sh` and `git diff --check` also pass.

### Collision check

- #12 covered syntactically malformed JSON that failed parsing.
- #37 covered a valid object root with `"mcpServers": null`, which is now normalized correctly.
- #177 re-reported malformed JSON but was closed as not reproducible because the syntax guard exists.

This case parses successfully, reaches the current post-validation code, and fails because the guard does not validate the root shape.

### Severity (your guess)

major

The optional registration aborts a healthy launcher and can block Monk tools in unrelated hosts until the user repairs or removes the Antigravity config. The file itself is preserved.

### OS

Windows 11 with Git Bash for the isolated POSIX regression; the vulnerable shell path is shipped for macOS and Linux.

### monkd version

Not involved; the failure happens after the local companion health check.

### Target cloud

None; network-free local fixture.

### Integration

### Fix PR

#265 adds the red-to-green regression and the validated fix across all three shipped launcher mirrors.

aikinley:
## Summary
- validate Antigravity's existing MCP config as a JSON object, not only as syntactically valid JSON
- preserve valid non-object configs and keep the healthy launcher running with a warning
- apply the fix to all three shipped launcher mirrors
- add a red-to-green regression for the Python fallback path

## Why
A valid JSON array passes the current syntax check, then the Python fallback calls `cfg.get(...)` and exits under `set -e`. The same incompatible shape also breaks the jq assignment path. A healthy monk-agent therefore appears to fail solely because an existing Antigravity config has the wrong root type.

## Validation
- `C:\Program Files\Git\bin\bash.exe tests/start-monk-agent-antigravity-nonobject-config.sh`
- `C:\Program Files\Git\bin\bash.exe -n scripts/start-monk-agent.sh`
- SHA-256 equality verified across all three launcher mirrors
- `git diff --check`

The regression fails on pristine v0.1.54 with `AttributeError: 'list' object has no attribute 'get'` and passes with this change while preserving the original config.

## Generated artifacts
`GENERATED.md` says the public tree is generated from the private `monk-agent/plugin` source. This PR updates the mirrored public artifacts consistently; the corresponding source change should be ported before the next regeneration.

Fixes #255

### DIFF
--- a/.antigravity-plugin/scripts/start-monk-agent.sh
+++ b/.antigravity-plugin/scripts/start-monk-agent.sh
@@ -106,14 +106,15 @@ register_antigravity_mcp() {
     # produce a warning, not abort the launcher after the health check passed.
     valid=1
     if command -v jq >/dev/null 2>&1; then
-      jq empty "$mcp_cfg" >/dev/null 2>&1 || valid=0
+      jq -e 'type == "object"' "$mcp_cfg" >/dev/null 2>&1 || valid=0
     elif command -v python3 >/dev/null 2>&1; then
       python3 -c 'import json, sys
 with open(sys.argv[1], encoding="utf-8") as handle:
-    json.load(handle)' "$mcp_cfg" >/dev/null 2>&1 || valid=0
+    cfg = json.load(handle)
+raise SystemExit(0 if isinstance(cfg, dict) else 1)' "$mcp_cfg" >/dev/null 2>&1 || valid=0
     fi
     if [ "$valid" = "0" ]; then
-      echo "Warning: $mcp_cfg is not valid JSON; skipping automatic Antigravity MCP registration." >&2
+      echo "Warning: $mcp_cfg is not a valid JSON object; skipping automatic Antigravity MCP registration." >&2
       printf '  Add manually if desired: {"mcpServers":{"monk":{"serverUrl":"%s"}}}\n' "$server_url" >&2
       return 0
     fi
--- a/plugins/monk/scripts/start-monk-agent.sh
+++ b/plugins/monk/scripts/start-monk-agent.sh
@@ -106,14 +106,15 @@ register_antigravity_mcp() {
     # produce a warning, not abort the launcher after the health check passed.
     valid=1
     if command -v jq >/dev/null 2>&1; then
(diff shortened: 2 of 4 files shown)


Lachann:

PR #267, opened 2026-08-05
fix: quote Antigravity config paths in uninstaller

### LINKED ISSUE #254: [Bug bounty] Antigravity uninstall silently skips cleanup for apostrophe home paths
### Stage

uninstall

### Coding agent

OpenAI Codex

### Coding agent version

### Repro steps

1. Check out v0.1.54 on a POSIX shell where `python3` is available and `jq` is absent, so `remove_antigravity_mcp` takes its Python fallback.
2. Set `HOME` to an isolated path containing an apostrophe, for example `.../home/O'Connor`.
3. Create `$HOME/.gemini/config/mcp_config.json` with both Monk and an unrelated server:

   ```json
   {
     "mcpServers": {
       "monk": {"serverUrl": "http://127.0.0.1:7419/mcp"},
       "preserved": {"serverUrl": "http://127.0.0.1:9999/mcp"}
     },
     "preserved": {"value": "still here"}
   }
   ```

4. Run `scripts/uninstall-monk-agent.sh --yes` with `MONK_AGENT_HOME` and `MONK_AGENT_INSTALL_DIR` also pointing inside that isolated home.
5. Inspect the config after the command reports completion.

I reproduced this with a network-free regression that isolates `HOME` and `PATH`, forces the Python branch, and asserts that only `mcpServers.monk` is removed. The current release fails deterministically:

```text
monk-agent uninstall complete.
Traceback (most recent call last):
  File "<string>", line 7, in <module>
AssertionError
```

The config is unchanged and still contains `mcpServers.monk`.

### Expected behavior

The official uninstaller removes only the Monk-owned MCP entry, preserves unrelated config, and reports completion.

### Actual behavior

`remove_antigravity_mcp` interpolates `$mcp_cfg` directly into generated Python source:

```python
with open('$mcp_cfg') as f:
...
with open('$mcp_cfg', 'w') as f:
```

The apostrophe terminates the Python string and produces a `SyntaxError`. The shell then swallows that failure with `2>/dev/null || true`, prints `monk-agent uninstall complete.`, and leaves a dead Monk MCP registration pointing at the removed companion.

I validated the minimal fix locally: pass the config path as a `sys.argv` value and use that variable in both `open` calls. The new regression then passes, as do `sh -n scripts/uninstall-monk-agent.sh` and `git diff --check`.

### Collision check

- #26 covered **install/startup registration** in `start-monk-agent.sh`; its v0.1.53 fix updated the three rendered launcher copies, not this uninstaller path.
- #27 covered uninstallers that had **no Antigravity cleanup at all** in v0.1.42; v0.1.50 added the cleanup function. This report exercises that shipped cleanup function's still-broken Python fallback in v0.1.53.

### Severity (your guess)

minor

The uninstall claims success but leaves persistent global host configuration behind. Later Antigravity sessions can keep trying to connect to a server the same uninstall just removed. Installing `jq` or avoiding an apostrophe in the home path works around it.

### OS

Windows 11 with Git Bash for the isolated POSIX regression; the vulnerable shell code is shipped for macOS and Linux.

### monkd version

Not involved; this is local plugin cleanup.

### Target cloud

None; network-free local source fixture.

### Integration

Google Antigravity MCP registration cleanup

### Fix PR

#267 adds the current-release red-to-green regression and the validated fix across both shipped POSIX uninstaller mirrors.

aikinley:
## Summary
- pass the Antigravity config path to the Python fallback through `sys.argv` instead of interpolating it into generated source
- keep removal scoped to `mcpServers.monk` while preserving unrelated configuration
- update both shipped POSIX uninstaller mirrors
- add a network-free regression for apostrophes in the config path

## Why
An apostrophe in `$HOME` terminates the generated Python string in v0.1.54. The fallback error is intentionally swallowed, so uninstall reports success but leaves the dead Monk MCP registration behind.

## Validation
- pristine v0.1.54 regression: exits 1 because `mcpServers.monk` remains
- fixed regression: `tests/uninstall-monk-agent-antigravity-python-path.sh` passes
- `C:\Program Files\Git\bin\bash.exe -n scripts/uninstall-monk-agent.sh`
- SHA-256 equality verified across both shipped uninstaller mirrors
- `git diff --check`

## Generated artifacts
`GENERATED.md` says the public tree is generated from the private `monk-agent/plugin` source. This PR updates the mirrored public artifacts consistently; the corresponding source change should be ported before the next regeneration.

Fixes #254

### DIFF
--- a/plugins/monk/scripts/uninstall-monk-agent.sh
+++ b/plugins/monk/scripts/uninstall-monk-agent.sh
@@ -194,9 +194,10 @@ remove_antigravity_mcp() {
     fi
   elif command -v python3 >/dev/null 2>&1; then
     python3 -c "
-import json
+import json, sys
+mcp_cfg = sys.argv[1]
 try:
-    with open('$mcp_cfg') as f:
+    with open(mcp_cfg) as f:
         cfg = json.load(f)
 except (OSError, ValueError):
     raise SystemExit(0)
@@ -205,10 +206,10 @@ if isinstance(servers, dict) and 'monk' in servers:
     del servers['monk']
     if not servers:
         cfg.pop('mcpServers', None)
-    with open('$mcp_cfg', 'w') as f:
+    with open(mcp_cfg, 'w') as f:
         json.dump(cfg, f, indent=2)
         f.write('\n')
-" 2>/dev/null || true
+" "$mcp_cfg" 2>/dev/null || true
   fi
 }
 
--- a/scripts/uninstall-monk-agent.sh
+++ b/scripts/uninstall-monk-agent.sh
@@ -194,9 +194,10 @@ remove_antigravity_mcp() {
     fi
   elif command -v python3 >/dev/null 2>&1; then
     python3 -c "
-import json
+import json, sys
+mcp_cfg = sys.argv[1]
 try:
-    with open('$mcp_cfg') as f:
+    with open(mcp_cfg) as f:
         cfg = json.load(f)
 except (OSError, ValueError):
     raise SystemExit(0)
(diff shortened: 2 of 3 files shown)


Lachann:

PR #374, opened 2026-08-29
fix: only nudge on explicit signed-out auth state

### LINKED ISSUE #364: [Bug bounty] POSIX launcher emits a false sign-out nudge for malformed or unrelated auth.json payloads
### Bounty eligibility

- [ ] I have used the installed Monk product for this specific finding (this report is from deterministic execution of the shipped v0.1.58 launcher path)

I am leaving the eligibility boxes unchecked rather than overstating bounty eligibility. The shipped launcher behaviour and regression are deterministic.

### Stage

connect

### Coding agent

OpenAI Codex

### Coding agent version

ChatGPT / Codex; Monk plugin v0.1.58 (`8d9a2c3e9b1fcf0ba185f010f9452b2162cd6b54`)

### Repro steps

`emit_signin_nudge` in the current POSIX launcher correctly strips JSON whitespace after #56, but then treats **every non-empty auth response that is not `signedIn:true` as signed out**:

```sh
compact_body="$(printf '%s' "$body" | tr -d '[:space:]')"
case "$compact_body" in
  *'"signedIn":true'*) return 0 ;;
esac
[ -n "$body" ] || return 0
# POST /plugin/nudge?type=signin ...
```

That contradicts the comment immediately above it, which says only an affirmative `signedIn:false` should reach the nudge.

I reproduced the full shipped launcher path with an isolated fake companion and curl fixture. The health probe returns the expected resource; only `/auth.json` varies. No account, cloud, or real runtime state is touched.

Current v0.1.58 results:

```text
malformed       body={"error":"keychain unavailable"}  nudges=1  false_message=1
missing         body={"status":"ok"}                   nudges=1  false_message=1
explicit_false  body={"signedIn":false}                nudges=1  false_message=1
explicit_true   body={"signedIn":true}                 nudges=0  false_message=0
```

So a non-auth object or a malformed/partial/error payload causes the launcher to POST `/plugin/nudge?type=signin` and inject:

```text
monk-agent is running but you are NOT signed in to Monk...
```

The PowerShell sibling already treats an unparseable JSON payload as non-definitive and returns without nudging.

### Expected behavior

The sign-in nudge should be emitted only when the response definitively reports `signedIn:false`. Empty, malformed, partial, or unrelated payloads should stay best-effort quiet rather than asserting that the user is signed out.

### Actual behavior

Any non-empty body that does not contain `"signedIn":true` is treated as a confirmed signed-out state. This can turn an auth endpoint error or unexpected response into a false re-authentication instruction and an unnecessary sign-in nudge.

### Duplicate check

This is distinct from #56. #56 covered valid `signedIn: true` JSON with insignificant whitespace being misclassified because the launcher used an exact compact substring. That whitespace fix is present on current main and passes. This report covers the remaining **absence of an explicit `signedIn:false` gate**: malformed/partial/unrelated non-empty responses still produce the false nudge.

I searched issues and PRs for malformed `auth.json`, invalid JSON, missing `signedIn`, sign-in nudge, and non-definitive auth payloads and found #56 only; no report covers this current v0.1.58 behaviour.

### Severity (your guess)

major

### OS

Linux/POSIX deterministic launcher reproduction. The affected POSIX launcher is also shipped for macOS.

### monkd version

Not involved; this is launcher interpretation of the local companion's `/auth.json` response.

### Target cloud

None.

### Integration

SessionStart launcher for supported POSIX hosts.

### Fix

A tested patch changes the decision to three states: explicit true => quiet, explicit false => nudge, anything else => quiet. The regression covers malformed, unrelated, explicit-false, and explicit-true responses and is red on current main / green with the patch.

Prepared with ChatGPT/Codex assistance.

paulcakeface:
Fixes #364.

The POSIX launcher previously treated every non-empty `/auth.json` body that was not `signedIn:true` as a confirmed signed-out state. That meant malformed, partial, or unrelated auth responses emitted the SessionStart sign-in warning and posted the sign-in nudge.

This change makes the shell path tri-state:
- `signedIn:true` -> quiet
- `signedIn:false` -> nudge
- anything else -> quiet

Regression coverage is added to the existing `start-monk-agent-fastpath.sh` test, which is already run by the repository's Ubuntu/macOS Install E2E job. The fixture exercises malformed, unrelated, explicit-false, and explicit-true payloads, and the test also asserts the three shipped POSIX launcher copies remain byte-identical.

Local validation:
- `./tests/start-monk-agent-fastpath.sh` ✅
- `./tests/start-monk-agent-readiness-timeout.sh` ✅
- `./tests/start-monk-agent-antigravity-python-path.sh` ✅
- `sh -n` on all three POSIX launcher copies ✅
- generated-copy `cmp` checks ✅
- `git diff --check` ✅

Prepared with ChatGPT/Codex assistance.

### DIFF
--- a/.antigravity-plugin/scripts/start-monk-agent.sh
+++ b/.antigravity-plugin/scripts/start-monk-agent.sh
@@ -214,12 +214,14 @@ emit_signin_nudge() {
   # JSON output may render as `"signedIn": true` with a space, which the exact
   # substring match below would otherwise miss and misreport as signed-out.
   compact_body="$(printf '%s' "$body" | tr -d '[:space:]')"
+  # Treat auth status as three-state: only an explicit signedIn:false is a
+  # confirmed signed-out result. signedIn:true is signed in; empty, malformed,
+  # partial, or unrelated payloads are non-definitive and must stay quiet.
   case "$compact_body" in
     *'"signedIn":true'*) return 0 ;;
+    *'"signedIn":false'*) ;;
+    *) return 0 ;;
   esac
-  # Empty body = read error / 500 / timeout, NOT a confirmed signed-out state —
-  # suppress the nudge. Only an affirmative signedIn:false reaches the nudge below.
-  [ -n "$body" ] || return 0
   # $client is resolved once at the top of the script (Cursor-aware ordering).
   nudge_url="$base_url/plugin/nudge?type=signin&client=$client"
   if command -v curl >/dev/null 2>&1; then
--- a/plugins/monk/scripts/start-monk-agent.sh
+++ b/plugins/monk/scripts/start-monk-agent.sh
@@ -214,12 +214,14 @@ emit_signin_nudge() {
   # JSON output may render as `"signedIn": true` with a space, which the exact
   # substring match below would otherwise miss and misreport as signed-out.
   compact_body="$(printf '%s' "$body" | tr -d '[:space:]')"
+  # Treat auth status as three-state: only an explicit signedIn:false is a
+  # confirmed signed-out result. signedIn:true is signed in; empty, malformed,
+  # partial, or unrelated payloads are non-definitive and must stay quiet.
   case "$compact_body" in
     *'"signedIn":true'*) return 0 ;;
+    *'"signedIn":false'*) ;;
+    *) return 0 ;;
   esac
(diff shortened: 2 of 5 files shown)


paulcakeface:
Bounty-deadline review nudge: this remains current against `main` and the focused launcher regression is green. If there’s anything you want changed before the 31 Aug submission cutoff, I can turn it around quickly.

PR #361, opened 2026-08-27
fix: bound POSIX installer download stalls

paulcakeface:
Fixes #360.

## What changed

- add configurable `MONK_AGENT_DOWNLOAD_CONNECT_TIMEOUT` (default 10s)
- add configurable `MONK_AGENT_DOWNLOAD_STALL_TIMEOUT` (default 30s)
- apply curl connect + low-speed/no-progress deadlines to both checksum and archive downloads
- bound wget fallback transfers to one timeout-limited attempt
- reject invalid/non-positive timeout values early
- keep all three shipped POSIX `ensure-monk-agent.sh` copies in sync
- add a deterministic regression test covering both curl download calls and the wget single-attempt invariant

## Why

The POSIX bootstrap previously delegated checksum/archive transfer duration entirely to curl/wget defaults. A server that accepts a connection but stops making progress can therefore hold plugin installation before the launcher reaches its separate readiness deadline.

This is distinct from #52 (post-launch readiness timeout) and #90 (offline fallback to a previously verified local agent).

## Validation

- `sh -n` on the patched installer and regression script
- deterministic regression: `ensure_monk_agent_download_timeout=pass calls=2`
- regression checks that the three shipped POSIX installer copies are byte-identical
- base verified against current upstream `main` at `8d9a2c3e9b1fcf0ba185f010f9452b2162cd6b54`

I also prepared Install E2E workflow wiring locally, but did not include it because the connected GitHub token does not have permission to modify `.github/workflows/*` on the fork. The regression itself is included and can be run directly with:

```sh
sh ./tests/ensure-monk-agent-download-timeout.sh
```

Prepared with ChatGPT/Codex assistance.

Issue #360, opened 2026-08-27
[Bug bounty] POSIX installer downloads have no stall deadline and can outlive SessionStart


paulcakeface:
### Bounty eligibility

- [x] I have signed up at monk.io with this GitHub account
- [x] I have used the product (installed the plugin and ran it, rather than only reading the code)
- [x] I have starred this repo

### Stage

install

### Coding agent

OpenAI Codex

### Coding agent version

ChatGPT Work / Codex; Monk plugin v0.1.58 (`8d9a2c3e9b1fcf0ba185f010f9452b2162cd6b54`)

### Repro steps

1. Check out plugin v0.1.58 on Linux or macOS and inspect/run the shipped POSIX bootstrap, `scripts/ensure-monk-agent.sh`, from a clean install directory.
2. Point `MONK_AGENT_DOWNLOAD_BASE` at an HTTP endpoint that accepts the checksum or archive request but then stops making meaningful progress.
3. Run the installer, for example:

```sh
MONK_AGENT_INSTALL_DIR="$(mktemp -d)" \
MONK_AGENT_DOWNLOAD_BASE=http://127.0.0.1:8080/stable \
  ./scripts/ensure-monk-agent.sh
```

4. The v0.1.58 checksum and archive paths invoke curl as `curl -fL ...` with no connect/no-progress deadline. The wget fallback likewise has no timeout/retry bound.
5. The bootstrap therefore inherits downloader/network defaults and has no plugin-owned deadline for either transfer. A stalled download can hold the install path before the launcher ever reaches its already-bounded readiness phase.

I also built a deterministic regression around the shipped script: a local curl fixture allows the checksum request, records the arguments supplied to both requests, and simulates curl exit 28 on the archive transfer. On the fixed script both checksum and archive calls carry a connect deadline plus a low-speed/no-progress deadline; the three shipped POSIX installer copies are asserted identical.

Duplicate check: this is distinct from #52, which covered the launcher readiness loop *after* the companion process has been started, and from #90, which covered update-service failure when a previously verified local agent is available. This report is the installer download itself having no bounded stall policy.

### Expected behavior

The installer should bound connection setup and periods with no meaningful download progress, while still allowing a slow but advancing agent archive to complete. The wget fallback should not multiply the bound through repeated retries. The limits should be configurable for unusual networks.

### Actual behavior

[...]

PR #376, opened 2026-08-29
fix: bound Windows agent download stalls

### LINKED ISSUE #375: [Bug bounty] Windows PowerShell installer downloads have no stall deadline and can outlive SessionStart
### Bounty eligibility

- [x] I have used the installed Monk product end-to-end through its authenticated MCP/runtime/deploy path

### Stage

install

### Coding agent

OpenAI ChatGPT / Codex-assisted testing

### Coding agent version

Monk plugin `v0.1.58` (`8d9a2c3e9b1fcf0ba185f010f9452b2162cd6b54`); exact shipped `scripts/ensure-monk-agent.ps1`

### Repro steps

The Windows installer has two blocking download calls:

```powershell
Invoke-WebRequest -Uri $ChecksumUrl -OutFile $ChecksumTmp -UseBasicParsing
...
Invoke-WebRequest -Uri $Url -OutFile $ArchiveTmp -UseBasicParsing
```

Neither call has a request timeout or a no-progress/stall deadline.

I reproduced this deterministically against the exact v0.1.58 PowerShell script with a local TCP/HTTP fixture and a fresh install directory. The fixture never contacts Monk or any third-party service.

#### Case 1: checksum connection succeeds, then no response arrives

1. Point `MONK_AGENT_DOWNLOAD_BASE` at a loopback listener.
2. The listener accepts the checksum TCP connection but sends no HTTP response bytes.
3. Run the shipped `ensure-monk-agent.ps1`.
4. The installer remains blocked until an **external 15-second safety cap** terminates it. It emits no internal timeout/error first.

Observed:

```text
exit=124 elapsed=15s
server: accepted checksum connection
stdout: <empty>
stderr: <empty>
```

#### Case 2: checksum succeeds, archive starts, then makes zero progress

1. The fixture returns a syntactically valid `.sha256` response immediately.
2. For the archive request it returns HTTP 200, a `Content-Length`, and the first **3 bytes** of the body.
3. It then sends nothing further.
4. The installer prints its normal install message and remains blocked until the same external 15-second cap kills it.

Observed:

```text
exit=124 elapsed=15s
accepted checksum
accepted archive
archive stalled after 3 bytes
Installing monk-agent from http://127.0.0.1:18081/windows/monk-agent-windows-latest.zip
```

This second case is the important one: the connection and response headers are already successful, so a connect-only deadline is insufficient. The file stream itself needs a bounded period of zero progress.

The deterministic harness used Microsoft PowerShell to exercise the exact shipped `.ps1` implementation. The affected production path is the Windows PowerShell bootstrap; the defect is script-owned timeout behaviour rather than an OS-specific remote endpoint failure.

### Expected behavior

The Windows installer should bound:

- connection / initial response setup; and
- periods with **no meaningful download progress** for both checksum and archive transfers.

A slow but continuously advancing 214 MB Windows agent archive should still be allowed to complete, so an absolute short transfer deadline is not ideal. The limits should be configurable for unusual networks.

### Actual behavior

Both Windows download calls inherit `Invoke-WebRequest` / network defaults with no plugin-owned stall policy. A server, proxy, captive portal, or broken distribution path that accepts the request and then stops sending bytes can hold the blocking plugin bootstrap indefinitely and consume the host SessionStart budget before the launcher reaches its already-bounded readiness handling.

This also weakens the existing offline/update fallback from #90: the checksum download is inside a `try/catch`, but a stalled request never returns to that catch, so the previously verified local binary fallback cannot run.

### Validation of the prepared fix

I prepared a companion fix that gives the PowerShell installer separate configurable defaults:

- `MONK_AGENT_DOWNLOAD_CONNECT_TIMEOUT=10`
- `MONK_AGENT_DOWNLOAD_STALL_TIMEOUT=30`

The download helper uses a raw `HttpWebRequest` file stream with `Timeout` plus `ReadWriteTimeout`, preserving Windows PowerShell 5.1 compatibility and avoiding an absolute deadline on a slow-but-progressing body.

Deterministic regression result:

```text
untouched v0.1.58: FAIL — installer exceeded its configured download deadline
fixed tree:          PASS — header_stall=bounded body_stall=bounded copies=3
```

Additional controls passed:

- invalid zero timeout fails closed with exit 2;
- all three distributed PowerShell installer copies are byte-identical;
- all changed PowerShell files parse cleanly;
- a complete local checksum/archive still downloads, verifies, extracts and installs;
- the patched script successfully downloaded and installed the **real current stable Windows agent** from `https://get.monk.io/stable/windows/...` in an isolated smoke test (about 9 seconds).

Exact prepared fix commit: `df69d4a060f391a57ca40969e935c703b22d663c`.

### Duplicate check

Fresh open/closed issue search immediately before filing covered Windows/PowerShell installer stalls, `Invoke-WebRequest`, checksum/archive timeouts, no-progress downloads and SessionStart hangs.

Closest reports are distinct:

- **#360** is my POSIX `.sh` installer report and fix. This issue is the independent PowerShell implementation (`ensure-monk-agent.ps1`), just as previous Windows/POSIX bootstrap defects in this repo have required fixes in separate code paths.
- **#195** covers Windows PowerShell 5.1's Internet Explorer parser dependency (`-UseBasicParsing`), not stalled transfers.
- **#90** covers update-service failure when a verified local binary exists; here a stalled checksum request never returns to the fallback catch.
- launcher readiness timeout issues occur after bootstrap; this defect blocks inside the download phase before agent startup/readiness.

No matching Windows PowerShell download-stall report was found.

### Severity (my guess)

major

A stalled distribution/proxy response can block the Windows plugin bootstrap indefinitely and prevent the Monk MCP companion from becoming available for the session.

### OS

Windows PowerShell bootstrap path. Deterministic reproduction executed the exact shipped `.ps1` in an isolated Microsoft PowerShell runtime; no cloud or third-party target was involved.

### monkd version

Not involved; the failure occurs while downloading `monk-agent.exe`, before the Monk runtime is required.

### Target cloud

None — local companion bootstrap.

### Integration

Official Monk Windows PowerShell companion installer.

paulcakeface:
Fixes #375.

The Windows PowerShell bootstrap currently uses unbounded `Invoke-WebRequest` file downloads. A server/proxy can accept the request (or send response headers and a few archive bytes) and then stop making progress, leaving the blocking SessionStart installer waiting indefinitely.

This change:

- adds configurable `MONK_AGENT_DOWNLOAD_CONNECT_TIMEOUT` (default 10s) and `MONK_AGENT_DOWNLOAD_STALL_TIMEOUT` (default 30s);
- validates both values as positive integer seconds;
- uses a Windows PowerShell 5.1-compatible `HttpWebRequest` streaming helper with separate request and read/write deadlines, so a slow-but-progressing archive is not subject to a short absolute transfer cap;
- preserves the existing raw-download behaviour that avoids the Windows PowerShell IE parser dependency;
- applies the change identically to all three distributed PowerShell installer copies;
- adds deterministic regression coverage for both a response-header stall and an archive that sends three bytes then stops;
- runs that regression in the existing Windows install E2E job.

Validation performed before opening:

- regression **fails** against untouched v0.1.58 by exceeding its configured deadline;
- regression **passes** on this branch: `header_stall=bounded body_stall=bounded copies=3`;
- invalid zero timeout exits 2;
- all changed PowerShell files parse cleanly;
- three installer copies are byte-identical;
- complete local checksum/archive install control passes;
- isolated smoke test against the real current stable `get.monk.io` Windows artifact completed successfully in ~9s.

This is the independent Windows PowerShell counterpart to #360 / #361, which only changes the POSIX shell installer.

### DIFF
--- a/.antigravity-plugin/scripts/ensure-monk-agent.ps1
+++ b/.antigravity-plugin/scripts/ensure-monk-agent.ps1
@@ -32,6 +32,53 @@ $DownloadBase = if ($env:MONK_AGENT_DOWNLOAD_BASE) { $env:MONK_AGENT_DOWNLOAD_BA
 }
 $AutoUpdate = if ($env:MONK_AGENT_AUTO_UPDATE) { $env:MONK_AGENT_AUTO_UPDATE } else { "1" }
 
+function Get-PositiveTimeoutSeconds {
+  param([string]$Name, [int]$DefaultValue)
+  $Raw = [Environment]::GetEnvironmentVariable($Name)
+  if ([string]::IsNullOrWhiteSpace($Raw)) {
+    return $DefaultValue
+  }
+  $Parsed = 0
+  if (-not [int]::TryParse($Raw, [ref]$Parsed) -or $Parsed -le 0) {
+    [Console]::Error.WriteLine("$Name must be a positive integer number of seconds.")
+    exit 2
+  }
+  return $Parsed
+}
+
+$DownloadConnectTimeout = Get-PositiveTimeoutSeconds "MONK_AGENT_DOWNLOAD_CONNECT_TIMEOUT" 10
+$DownloadStallTimeout = Get-PositiveTimeoutSeconds "MONK_AGENT_DOWNLOAD_STALL_TIMEOUT" 30
+
+# Windows PowerShell's Invoke-WebRequest has no no-progress timeout for a file
+# body. A distribution endpoint can accept the connection (or send headers and a
+# few bytes) and then hold the blocking SessionStart installer indefinitely.
+# HttpWebRequest exposes separate request and stream read/write deadlines on
+# Windows PowerShell 5.1. ReadWriteTimeout resets on successful stream I/O, so a
+# slow but advancing archive is not subject to an absolute transfer deadline.
+function Invoke-BoundedDownload {
+  param([string]$Uri, [string]$OutFile)
+  $Request = [System.Net.HttpWebRequest]::Create($Uri)
+  $Request.Timeout = $DownloadConnectTimeout * 1000
+  $Request.ReadWriteTimeout = $DownloadStallTimeout * 1000
+  $Request.AllowAutoRedirect = $true
+  $Response = $null
+  $Input = $null
+  $Output = $null
+  try {
+    $Response = $Request.GetResponse()
+    $Input = $Response.GetResponseStream()
+    $Output = [System.IO.File]::Open($OutFile, [System.IO.FileMode]::Create, [System.IO.FileAccess]::Write, [System.IO.FileShare]::None)
+    $Buffer = New-Object byte[] 65536
+    while (($Read = $Input.Read($Buffer, 0, $Buffer.Length)) -gt 0) {
+      $Output.Write($Buffer, 0, $Read)
+    }
+  } finally {
+    if ($Output) { $Output.Dispose() }
+    if ($Input) { $Input.Dispose() }
+    if ($Response) { $Response.Dispose() }
+  }
+}
+
 $Target = Join-Path $InstallDir "monk-agent.exe"
 $ChecksumInstalled = Join-Path $InstallDir "monk-agent.sha256"
 $MonkHome = if ($env:MONK_AGENT_HOME) { $env:MONK_AGENT_HOME } else { Join-Path $HOME ".monk" }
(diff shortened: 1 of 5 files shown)


paulcakeface:
Bounty-deadline review nudge: the Windows download-stall fix is current against `main`; deterministic header/mid-archive stall regressions and the real stable-download smoke test are green. Happy to adjust quickly if you want anything changed before the 31 Aug cutoff.

PR #492, opened 2026-09-03
fix(antigravity): kill stale monk-agent PID in ensure hooks (#394)

### LINKED ISSUE #394: [Bug bounty] Ensure hooks spawn monk-agent without killing stale PID when health probe fails
## Bounty eligibility

- [x] I have signed up at monk.io with this GitHub account (NyxSpecter4)
- [x] I used the installed Monk product end-to-end against a real disposable app for this reproduction

## Stage

connect

## Coding agent

Cursor

## Coding agent version

1.7.33

## Summary

When the monk-agent health probe fails (e.g. hung process still holding :7419 per #378 class), `scripts/start-monk-agent.{ps1,sh}` kill the old PID via `Stop-ManagedAgent` before starting a new agent. The PreInvocation ensure hooks (`.antigravity-plugin/hooks/ensure-monk-agent.{ps1,sh}`) start a new process and overwrite the PID file **without** reading or killing the stale process. This can leave hung orphans invisible while the hook loop reports "did not become ready".

## Repro steps

1. Run static asymmetry fixture:

```powershell
powershell -ExecutionPolicy Bypass -File .\tools\ensure-hook-no-kill-fixture.ps1
```

2. Compare sources (plugin v0.1.58 cache):
   - `scripts/start-monk-agent.ps1` — `Stop-ManagedAgent` kills old PID (L99-122)
   - `.antigravity-plugin/hooks/ensure-monk-agent.ps1` — `Start-Process` + overwrite PID file, no kill (L97-129)
   - Same asymmetry on `.sh` siblings

## Expected behavior

Ensure hooks use the same stale-PID cleanup as `start-monk-agent` before spawning a replacement agent.

## Actual behavior

Static source analysis + fixture exit 0 **ASYMMETRY CONFIRMED**: start scripts kill old PID; ensure hooks do not. Related to #378 RPC wedge class but distinct surface (PID hygiene on ensure path).

## Severity (your guess)

major

## OS

Windows 11

## monkd version

v3.21.3 (monk-agent v0.1.58)

## Target cloud

local

## Not a duplicate

Not #362/#382 (block-monk hang). Not #286 (WSL relay reboot). No open sibling for ensure-hook no-kill asymmetry.

Evidence attached in comment (source excerpts + fixture output).

ZacLou:
Closes #394

The PreInvocation ensure hooks (`.antigravity-plugin/hooks/ensure-monk-agent.ps1` and `.sh`) now kill a stale `monk-agent` process recorded in their PID file before spawning a replacement, matching the cleanup already performed by `scripts/start-monk-agent.{ps1,sh}`.

- PowerShell: adds `Stop-StaleAgent` and calls it after the PID file path is resolved.
- POSIX shell: adds `stop_stale_agent` and calls it before starting the binary.
- Adds regression tests for both paths.

### DIFF
--- a/.antigravity-plugin/hooks/ensure-monk-agent.ps1
+++ b/.antigravity-plugin/hooks/ensure-monk-agent.ps1
@@ -92,6 +92,48 @@ function Test-AgentRunning {
   }
 }
 
+# Kill a stale monk-agent process recorded in the PID file before starting a
+# replacement. Mirrors Stop-ManagedAgent in scripts/start-monk-agent.ps1 so the
+# ensure path does not leave hung orphans behind (plugin#394).
+function Stop-StaleAgent {
+  param([string]$PidFilePath, [string]$ExpectedPath)
+
+  if (-not (Test-Path $PidFilePath)) { return }
+
+  $RawPid = (Get-Content -Raw $PidFilePath).Trim()
+  if (-not $RawPid) { return }
+
+  $ParsedPid = 0
+  if (-not [int]::TryParse($RawPid, [ref]$ParsedPid) -or $ParsedPid -le 0) {
+    Remove-Item -Force $PidFilePath -ErrorAction SilentlyContinue
+    return
+  }
+
+  $OldProcess = Get-Process -Id $ParsedPid -ErrorAction SilentlyContinue
+  if (-not $OldProcess) { return }
+
+  $ProcessPath = ""
+  try { $ProcessPath = $OldProcess.Path } catch { $ProcessPath = "" }
+
+  $KillIt = $false
+  if ($ProcessPath) {
+    try {
+      $ExpectedFull = [IO.Path]::GetFullPath($ExpectedPath)
+      $ProcessFull = [IO.Path]::GetFullPath($ProcessPath)
+      if ($ExpectedFull -ieq $ProcessFull) { $KillIt = $true }
+    } catch { }
+  } elseif (Test-Path $ExpectedPath) {
+    # No path available from the OS; only kill if the expected binary exists
+    # as a safety guard against killing an unrelated process that reused the PID.
+    $KillIt = $true
+  }
+
+  if ($KillIt) {
+    Stop-Process -Id $OldProcess.Id -Force -ErrorAction SilentlyContinue
+    Wait-Process -Id $OldProcess.Id -Timeout 5 -ErrorAction SilentlyContinue
+  }
+}
+
 # Fast path - already up. No telemetry here: this is a PreInvocation hook that
 # fires per model step, so emitting on the warm path would spam. The beacon fires
 # only on the cold-start paths below (install-needed / (re)start), which is the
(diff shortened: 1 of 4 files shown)


NyxSpecter4:
Reporter of #394 here — verified: the ensure hooks now mirror `Stop-ManagedAgent` (table on #394). The PID-reuse guard (`/proc/<pid>/exe` on Linux, `ps -o comm=` elsewhere, `$OldProcess.Path` on Windows) is the right shape; thanks for not killing a bare PID.

Two notes:
1. `tests/ensure-monk-agent-kill-stale-pid.sh` on Linux (Ubuntu, `sh` = dash): the stale-kill assertion holds, but the final check fails with `replacement monk-agent process (<pid>) is not running`. Cause is in the harness, not the fix: the fake `monk-agent` is a copy of `sh`, so `monk-agent serve --host …` tries to open a file named `serve` in the hook's cwd, not the `serve` stub on `PATH`. A one-line wrapper script that `exec sleep`s regardless of args fixes the test.
2. `Stop-StaleAgent` `elseif (Test-Path $ExpectedPath) { $KillIt = $true }` — when the OS gives no `Path` (access denied / 32-vs-64-bit), this kills the PID as long as the expected binary exists on disk, which is almost always. Consider `$OldProcess.ProcessName -like 'monk-agent*'` as the fallback instead, matching the `.sh` side's `comm=` check.

Branch is based on v0.1.59; `main` is v0.1.60 — rebase to keep the diff to 4 files.

ZacLou:
@NyxSpecter4 thanks for the review. Pushed both fixes:\n\n1. The shell test now runs the hook from `$fake_bin` so the fake binary (copy of sh) finds the `serve` stub in its cwd and the replacement process stays alive.\n2. `Stop-StaleAgent` now falls back to `$OldProcess.ProcessName -like 'monk-agent*'` instead of `Test-Path $ExpectedPath`, matching the POSIX side's `comm=` check and avoiding killing unrelated reused PIDs.\n\nPlease take another look.

NyxSpecter4 (approved):
As the reporter of #394, I took another look at the updated head (`7cd3668`):

1. **Shell regression test** — ran `tests/ensure-monk-agent-kill-stale-pid.sh` on Ubuntu (`sh` = dash): passed. Stale process killed, replacement recorded in the PID file and running. The `fake_bin` cwd fix resolves the harness issue I flagged.
2. **PowerShell** — code review only, since I have no Windows runner here: `Stop-StaleAgent` now falls back to `$OldProcess.ProcessName -like 'monk-agent*'` when the OS provides no process path, matching the POSIX side's `comm=` guard. A bare PID is never killed. I could not execute the `.ps1` test in this environment, so that one remains unverified by me.
3. **Rebase** — branch is based on `v0.1.60` (`8d8efce`); the diff is exactly the four hook/test files (356 insertions, no unrelated changes).

Both notes from my earlier review are addressed, and the PID-reuse guard is preserved on both platforms. Approving from the reporter's side — merging is of course up to the maintainers. Thanks for the quick turnaround on both points.

NyxSpecter4:
Both fixes verified: `Stop-StaleAgent` now falls back to `ProcessName -like 'monk-agent*'` only when no path is available (a reused PID is never killed unless the name matches — consistent with the POSIX `comm=` guard), and the `$fake_bin` cwd fix resolves the harness failure.

Ran `tests/ensure-monk-agent-kill-stale-pid.sh` on Linux/dash: **exit 0, "stale PID kill test passed"**. LGTM — good to merge.

NyxSpecter4 (approved):
Approving: both requested fixes verified in code; `tests/ensure-monk-agent-kill-stale-pid.sh` run on Linux/dash -> exit 0, "stale PID kill test passed". Note: .ps1 regression statically reviewed only (no pwsh available here); Stop-StaleAgent path-match + name-fallback logic is sound.

